PR #30, opened 2022-03-08
UI/mainscreen skeleton

Oxydeas:
This is the basic implementation of the main screen of the app with three functional buttons


Oxydeas:
Code coverage is low because there is still the MainActivity from the template which I do not use

sansive:
> Code coverage is low because there is still the MainActivity from the template which I do not use

Maybe we can delete the MainActivity as we are not going to use it and it was only part of the bootcamp

app/src/main/java/com/github/houseorganizer/houseorganizer/MainActivity.java
```diff
@@ -19,15 +19,14 @@ protected void onCreate(Bundle savedInstanceState) {
...
         // Get name
         EditText editText = findViewById(R.id.mainName);
         String message = editText.getText().toString();
 
         // An intent is an object that provides runtime binding between separate components
         // The intent represents an app's intent to do something
-        Intent intent = new Intent(this, GreetingActivity.class);
+        Intent intent = new Intent(this, MainScreenActivity.class);
         // An intent can carry data types as key-value pairs called extras
         intent.putExtra(EXTRA_MESSAGE, message);
```
Geno22, line 31: As the MainScreenActivity does not require a message, it should be better to entirely remove the message sending to not be confused later.

Geno22 (approved):
The main screen is good not a big fan of the icons but that is secondary and could be changed later on.

app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_main_screen);
+    }
+
+    public void houseButtonPressed(View view){
+        TextView text = findViewById(R.id.last_button_activated);
+        String s = "House button pressed";
```
Arkadyus, line 19: (Question) It might be better to store this string as a resource string in res/values/strings.xml (same for the other two strings), or maybe it's better if we keep them here since they won't be part of the end product (easier to delete) ?

app/src/main/res/values/strings.xml
```diff
@@ -1,5 +1,9 @@
 <resources>
     <string name="app_name">House Organizer</string>
+    <string name="settings_speakable_text">Settings</string>
+    <string name="house_speakable_text">Home selection</string>
+    <string name="info_speakable_text">Info</string>
+    <string name="button_feedback_text">No button has been pressed</string>
     <string name="mainNameHint">Name</string>
```
Arkadyus, line 7: (Question) What do app_name and mainNameHint do ? 
(Notation) mainNameHint should probably be main_name_hint to follow the notation

Arkadyus (approved):
Everything seems fine, just one or two things that might need to be cleaned but nothing important

app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
Oxydeas: As this is only here for testing purposes and probably be gone next week, I think we can keep them here so we don't have to modify multiple files when getting rid of it.

app/src/main/res/values/strings.xml
Oxydeas: mainNainHint comes from the template. I will delete it now as well as other useless strings.
App name is the name of the app display at the top

PR #31, opened 2022-03-09
Origin/feature/calendar

Arkadyus:
This pull request adds the calendar widget on the main screen along with a button that rotates between 3 different calendar views : one that shows every day of the month, one for days of the week and one that shows upcoming events in a list. 
The monthly and weekly views arent as complete as the upcoming one because they feel too dense/would take quite some time to implement further.


aindreias (dismissed):
Some of the code uses API level 26 / 24 features, while our minimum required API level is 23. In order for the tests to complete, these API calls need to be replaced with something <=23 .

aindreias (approved):
The changes since my last review look good to me

sansive (approved):
Nice work!

PR #32, opened 2022-03-09
Feature/login backend

Geno22:
Implementation of a basic login UI and the Google sign-in along with Firebase.
Parts need to be adapted to the new MainScreen activity so are blank for now.


app/build.gradle
```diff
@@ -33,9 +34,19 @@ android {
 
 dependencies {
 
+    // Import the BoM for the Firebase platform
+    implementation platform('com.google.firebase:firebase-bom:29.1.0')
+
     implementation 'androidx.appcompat:appcompat:1.4.1'
     implementation 'com.google.android.material:material:1.5.0'
     implementation 'androidx.constraintlayout:constraintlayout:2.1.3'
+    implementation 'com.google.firebase:firebase-auth:21.0.1'
```
aindreias, line 43: Since we're using the Firebase BoM, I think you can safely delete this line (we would still have the version-less firebase-auth dependency below)
```suggestion
```

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/LoginActivityTest.java
```diff
@@ -17,22 +17,11 @@
...
+public class LoginActivityTest {
 
     @Rule
-    public ActivityScenarioRule<MainActivity> testRule = new ActivityScenarioRule<>(MainActivity.class);
+    public ActivityScenarioRule<LoginActivity> testRule = new ActivityScenarioRule<>(LoginActivity.class);
+    
 
-    // Test that when the button is clicked an intent is fired and the intent contains the username
-    @Test
-    public void clickButton() {
```
Defibrilator, line 27: I would suggest to leave the test I added since it tests if the sign in button works

`// Test that when the sign in button is clicked an intent is fired
    @Test
    public void clickSignInButtonFiresIntent() {
        Intents.init();
        onView(withId(R.id.sign_in_button)).perform(click());
        intended(toPackage("com.github.houseorganizer.houseorganizer"));
        Intents.release();
    }`

aindreias (approved):
The code looks good to me. However, codeclimate seems to complain about a complexity issue for LoginActivity::onStart. You might need to fix it before merging.

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/LoginActivityTest.java
Geno22: The thing is it does not pass anymore as there is no intent fired when the sign-in is done for now. It will be useful again once we implement the compatibility with the main activity.

Defibrilator (approved):

Arkadyus (commented):
Pulling main on the branch would help merge the activities together easier and that way the coverage from codeclimate will be right and wont give problems.
Other than that the code looks good to me.

app/src/main/AndroidManifest.xml
```diff
@@ -18,6 +18,16 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        <activity
+            android:name=".LoginActivity"
+            android:exported="true" />
+        <activity
```
aindreias, line 24: I think we should delete the dangling reference to MainActivity

PR #34, opened 2022-03-10
Feature/house selection

sansive:
## Task
Add house selection activity

## Testing
* UI tests
* Small test changed for the main screen as the house button now launches the house selection activity


app/src/main/java/com/github/houseorganizer/houseorganizer/HouseSelectionActivity.java
```diff
@@ -0,0 +1,34 @@
...
+import android.view.View;
+
+import androidx.appcompat.app.AppCompatActivity;
+import androidx.recyclerview.widget.LinearLayoutManager;
+import androidx.recyclerview.widget.RecyclerView;
+
+public class HouseSelectionActivity extends AppCompatActivity {
+
+    RecyclerView housesView;
+    final String[] houseNames = {"House 1", "House 2", "House 3", "House 4"};
```
Oxydeas, line 14: We might want to customize house names in the future. But as this is only a place holder I think this is fine.

Oxydeas (approved):

app/src/main/java/com/github/houseorganizer/houseorganizer/HouseSelectionActivity.java
aindreias: In the future, we could to ask the user to pick a nickname for each household. In fact, this is how I structured the dummy firestore database, for now.

aindreias (approved):

app/src/main/java/com/github/houseorganizer/houseorganizer/HouseSelectionActivity.java
sansive: I agree, it will be done when the database is ready.

PR #58, opened 2022-03-11
Bug/login

Defibrilator:
I fixed the login by adding a debug.keystore.jks to the root with a single key for everyone, linked it in the build.gradle and added the SHA1 to firebase, I also restructured the LoginActivity to make it a bit cleaner.


Oxydeas (approved):
Thank you for fixing this.

aindreias (approved):
Great job, thank you for the fix!

PR #60, opened 2022-03-16
Feature/calendar firebase

Arkadyus:
Lets the user add events to his household calendar (for now dummy events). Also lets the user refresh the household calendar in case the view is not up to date with the database.


aindreias (approved):
Overall great contribution. Refreshing the calendar data is a nice touch

PR #69, opened 2022-03-17
Feature/email login

Defibrilator:
Added the possibility to use email and password to create an account. Created two activities, LoginEmail and RegisterEmail for that purpose, inputs are checked so that the user cannot put invalid credentials but no verification email is sent, maybe a future task.


app/src/main/java/com/github/houseorganizer/houseorganizer/LoginActivity.java
```diff
@@ -40,8 +43,20 @@ protected void onCreate(Bundle savedInstanceState) {
...
+    public void onClick(View v) {
+        switch (v.getId()) {
+            case R.id.email_signin_button:
+                startActivity(new Intent(this, LoginEmail.class));
+                break;
+            case R.id.google_sign_in_button:
+                googleSignInResultLauncher.launch(new Intent(mGoogleSignInClient.getSignInIntent()));
+                break;
+        }
     }
```
Arkadyus, line 60: It might be better to use setOnClickListener for each view in onCreate, or to add an android:onClick in the xml instead of this (https://developer.android.com/guide/topics/ui/controls/button or MainScreenActivity have examples)

app/src/main/java/com/github/houseorganizer/houseorganizer/LoginEmail.java
```diff
@@ -0,0 +1,81 @@
...
+    public void onClick(View v) {
+        switch (v.getId()) {
+            case R.id.log_email_signin_button:
+                if (inputsNotEmpty()) signInWithEmail(v);
+                break;
+            case R.id.log_email_register_button:
+                startActivity(new Intent(LoginEmail.this, RegisterEmail.class));
+                break;
+        }
+    }
```
Arkadyus, line 43: Same as onClick in LoginActivity, unless there is a reason to write it this way it is cleaner to do it either in onCreate or in the xml

app/src/main/java/com/github/houseorganizer/houseorganizer/LoginEmail.java
```diff
@@ -0,0 +1,81 @@
...
+        mAuth.signInWithEmailAndPassword(email, password)
+                .addOnCompleteListener(this, task -> {
+                    if (task.isSuccessful()) {
+                        // Sign in success, update UI with the signed-in user's information
+                        Log.d(TAG, "signInWithEmail:success");
+                        startActivity(new Intent(LoginEmail.this, MainScreenActivity.class));
+                        finish();
+                    } else {
+                        // If sign in fails, display a message to the user.
+                        Log.d(TAG, "signInWithEmail:failure");
```
Arkadyus, line 76: Same TAG issue as in LoginActivity too

app/src/main/java/com/github/houseorganizer/houseorganizer/RegisterEmail.java
```diff
@@ -0,0 +1,100 @@
...
+        mAuth.createUserWithEmailAndPassword(email, password)
+                .addOnCompleteListener(this, task -> {
+                    if (task.isSuccessful()) {
+                        // Sign in success, update UI with the signed-in user's information
+                        Log.d(TAG, "createUserWithEmail:success");
+                        startActivity(new Intent(RegisterEmail.this, MainScreenActivity.class));
+                        finish();
+                    } else {
+                        // If sign in fails, display a message to the user.
+                        Log.w(TAG, "createUserWithEmail:failure", task.getException());
```
Arkadyus, line 95: same TAG issue again

Arkadyus (changes requested):
Good code and everything works well ! Just a little refactoring/cleaning and it'll be perfect imo

app/src/main/java/com/github/houseorganizer/houseorganizer/RegisterEmail.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_register_email);
+
+        mAuth = FirebaseAuth.getInstance();
+
+        findViewById(R.id.reg_email_register_button).setOnClickListener(
+                v -> {
+                    if (isValidEmail() && isValidPassword()) signUpWithEmail(v);
```
sansive, line 31: Shouldn't we check if the email has been previously used to create another account on the app?

app/src/main/java/com/github/houseorganizer/houseorganizer/RegisterEmail.java
```diff
@@ -0,0 +1,100 @@
...
+        String password2 = password2_field.getText().toString();
+
+        boolean samePasswords = password.equals(password2);
+
+        // Regex to check valid password.
+        String regex = "^(?=.*[0-9])"
+                + "(?=.*[a-z])(?=.*[A-Z])"
+                + "(?=.*[@#$%^&+=])"
+                + "(?=\\S+$).{8,20}$";
+        Pattern p = Pattern.compile(regex);
```
Geno22, line 67: The underscore character isn't in the regex list, but is used in many passwords.
Is it possible to add it ?

app/src/main/res/layout/activity_register_email.xml
```diff
@@ -0,0 +1,91 @@
...
+        android:layout_width="300dp"
+        android:layout_height="38dp"
+        android:textColor="#F44336"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintHorizontal_bias="0.495"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/reg_email_constraints_message"
+        app:layout_constraintVertical_bias="0.122" />
+</androidx.constraintlayout.widget.ConstraintLayout>
```
Geno22, line 91: I don't know if it is supposed to last but the error text is displayed at the bottom of the page which is not visible because of the android keyboard. I think placing it higher will be clearer

Geno22 (changes requested):

PR #87, opened 2022-03-18
Feature/virtual homes

sansive:
## Task
Implement logic behind the virtual homes with firebase

## Testing
Comment every test for MainScreenActivity and HouseSelectionActivity until the mock for firebase authenticator is done.


app/src/main/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivity.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    public void submitHouseholdToFirestore(View view){
+        TextView houseHoldNameView = findViewById(R.id.editTextHouseholdName);
+        CharSequence houseHoldName = houseHoldNameView.getText();
+
+        Map<String, Object> houseHold = new HashMap<>();
+        Map<String, Object> residents = new HashMap<>();
+
+        residents.put("0", Uid);
```
Geno22, line 41: I know nothing about firebase but is it possible to have a <Integer, Object> map ?

app/src/main/res/layout/activity_main_screen.xml
```diff
@@ -121,4 +121,17 @@
...
+        android:layout_height="wrap_content"
+        android:layout_marginTop="100dp"
+        android:layout_marginEnd="56dp"
+        android:onClick="addHouseholdButtonPressed"
+        android:text="@string/add_household"
+        app:layout_constraintEnd_toStartOf="@+id/refresh_calendar"
+        app:layout_constraintHorizontal_bias="0.0"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent" />
+
```
Geno22, line 136: The button is displayed a little on top of the calendar.
It is temporary I know but not pretty for the presentation ^^
ecranAddHouse

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/MainScreenActivityTest.java
```diff
@@ -202,5 +204,4 @@ public void calendarViewRotatesCorrectly() {
     }
 
     // TODO : Add more meaningful tests for each row in the RecyclerViews (no idea how to do it)
-
-}
+}*/
```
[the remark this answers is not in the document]
sansive: These tests will be uncommented when the mock for the firebase user is done, otherwise they always fail because the new code needs to know the user that is using the app.

app/src/main/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivity.java
```diff
@@ -0,0 +1,63 @@
...
+
+        db.collection("households").add(houseHold)
+                .addOnSuccessListener(documentReference -> {
+                    Toast.makeText(view.getContext(),
+                                   view.getContext().getString(R.string.add_success),
+                                   Toast.LENGTH_SHORT).show();
+                })
+                .addOnFailureListener(documentReference -> Toast.makeText(view.getContext(),
+                                      view.getContext().getString(R.string.add_fail),
+                                      Toast.LENGTH_SHORT).show());
```
Arkadyus, line 56: The strings used here are the ones from the event addition (the names should be more precise on these that is my bad).
Some new ones should probably be made for household addition

app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
```diff
@@ -42,9 +47,45 @@ protected void onCreate(Bundle savedInstanceState) {
...
+                                households.add(id);
+                            }
+
+                            currentHouse = db.collection("households").document(households.get(0));
+                            text.setText("currentHouse: " + currentHouse.getId() + " by default");
+
+                        } else {
+                            Toast.makeText(getApplicationContext(), "Could not get a house.", Toast.LENGTH_SHORT).show();
+                        }
+                    });*/
```
Arkadyus, line 83: Is the user not signed in by the time he gets on this activity ? We can probably make it so that if no houses are found fall back to the temp case for now

Arkadyus (changes requested):
Good implementation, just needs a few changes as said in the comments.
Also maybe the CreateHouseActivity should be just a dialog and not a full on activity maybe, but that can be changed later since this way works for now.

app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
sansive: For this activity the user should be signed in. If no houses are found that means that there was an error trying to retrieve the info from firebase, therefore I think showing an error message is best.

app/src/main/res/layout/activity_main_screen.xml
Oxydeas: I know it was just a placeholder for now. I will change it when merged

app/src/main/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivity.java
```diff
@@ -0,0 +1,63 @@
...
+import android.widget.Toast;
+
+import com.google.firebase.firestore.FirebaseFirestore;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
```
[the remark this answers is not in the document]
Oxydeas: Done

app/src/main/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivity.java
Oxydeas: For the comment on the residents map, this is the way I found to immitate an array on firebase. Basically a map from index to the object we want; and also it works

Oxydeas: Strings fixed

app/src/main/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivity.java
Oxydeas: It is a string, not an integer ;)

Oxydeas (approved):
I reviewed the files again and it seems that everything has been said for now. Still have the testing part and multiple conflicts with the main branch but it is ready for merging.

PR #88, opened 2022-03-22
only display upcoming events + clean refreshing when adding

Arkadyus:
The calendar now correctly only displays events that are in the future, addEvent also calls refreshCalendar instead of pseudo-refreshing it itself

### DIFF
--- a/app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
+++ b/app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
@@ -56,6 +56,7 @@ public class MainScreenActivity extends AppCompatActivity {
         findViewById(R.id.calendar_view_change).setOnClickListener(this::rotateView);
         findViewById(R.id.add_event).setOnClickListener(this::addEvent);
         findViewById(R.id.refresh_calendar).setOnClickListener(this::refreshCalendar);
+        refreshCalendar(findViewById(R.id.calendar));
     }
 
     private void signOut(View v) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
```diff
@@ -78,27 +78,23 @@ private void rotateView(View v) {
 
     private void addEvent(View v) {
         Map<String, Object> data = new HashMap<>();
-        Event event = new Event("added", "this is the event that i added using the add button", LocalDateTime.now(), 100);
         data.put("title", "added");
         data.put("description", "this is the event that i added using the add button");
-        data.put("start", LocalDateTime.now().toEpochSecond(ZoneOffset.UTC));
+        data.put("start", LocalDateTime.now().plusHours(2).toEpochSecond(ZoneOffset.UTC));
```
Geno22, line 83: This is for testing purpose is that it ? Just to be sure : in the future the starting time will be decided by the user ?

Geno22 (approved):
It seems all good nice changes

app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
Arkadyus: yes thats what im implementing in the feature/event_editing branch this sprint

Oxydeas: We might also need to consider various timezones when adding events, our users will not all live in UTC

Oxydeas (approved):
Godd small change. Cirrus still complains about useless resources, we should do a cleanup branch next week

app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
Arkadyus: I dont think it is a problem because all we do is store it using UTC and then load it back from UTC so the part where it uses UTC is the backend from how i understand it.

PR #89, opened 2022-03-23
Feature/event editing

Arkadyus:
Added the ability to edit the fields of the event the user is adding
Added the ability to edit or delete an event from the household upon clicking on its title
Both implementations work although they are not tested because they need the database emulation that should be made in #75 
Since there is a lot of code, Codeclimate might complain so ill do my best to improve the code ASAP (I dont think I can run Codeclimate locally hence why im doing the PR now)


sansive (commented):
There are some warnings in the code due to concatenating strings inside method 'setText'. Instead of doing that you can use placeholders like this: 

```
<string name="welcome_messages">Hello, %1$s! You have %2$d new messages.</string>
getString(R.string.welcome_message, "Test", 0);
```

sansive (commented):
There are more warnings in the layout files: missing `autofillHints` attribute, missing accessibility label, ...

app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
```diff
@@ -116,21 +122,39 @@ private void rotateView(View v) {
...
+        final String title = ((EditText) dialogView.findViewById(R.id.new_event_title)).getText().toString();
+        final String desc = ((EditText) dialogView.findViewById(R.id.new_event_desc)).getText().toString();
+        final String date = ((EditText) dialogView.findViewById(R.id.new_event_date)).getText().toString();
+        final String duration = ((EditText) dialogView.findViewById(R.id.new_event_duration)).getText().toString();
+        Map<String, String> event = new HashMap<>();
+        event.put("title", title);
+        event.put("desc", desc);
+        event.put("date", date);
+        event.put("duration", duration);
+        if (Util.putEventStringsInData(event, data)) {
```
aindreias, line 146: How about making `Util::putEventStringsInData` return `false` in case of an error and true otherwise? This way you can change line 146 here to `if (! Util.putEventStringsInData(...))` which to me seems more intuitive.

aindreias (approved):
Overall good code, thank you for the contribution!
Calling dialog.dismiss() might be redundant in certain places, but we can test this next week.

sansive (approved):

PR #93, opened 2022-03-24
Feature/groceries list

Geno22:
Added data structure for the shopping list and a simple implementation on the MainScreen.


app/src/main/res/layout/activity_main_screen.xml
```diff
@@ -87,6 +87,18 @@
         app:layout_constraintTop_toBottomOf="@+id/house_imageButton"
         app:layout_constraintVertical_bias="0.064" />
 
+    <Button
+        android:id="@+id/liste_view_change2"
```
Arkadyus, line 91: list_view_change is probably cleaner unless it conflicts with any other id

Arkadyus (approved):
Everything looks clean, moving ShopItem inside of ShopList can be refactored later as it is not that big of an issue and is quickly fixed.

app/src/main/java/com/github/houseorganizer/houseorganizer/ShopListAdapter.java
```diff
@@ -0,0 +1,58 @@
...
+import android.view.ViewGroup;
+import android.widget.CheckBox;
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView;
+
+public class ShopListAdapter extends RecyclerView.Adapter<ShopListAdapter.ItemsHolder> {
+
+    public class ItemsHolder extends RecyclerView.ViewHolder{
+        public CheckBox checkBox;
```
aindreias, line 15: Great idea to use a `CheckBox` button! I will probably switch the simple task buttons to this as well.

aindreias (approved):
I like the idea of switching between task and grocery lists. Overall good job and thanks for adding unit tests.

PR #95, opened 2022-03-24
Add basic anon sign-in

aindreias:
New or signed-out users can now browse the app with an anonymous firebase account by clicking the `Discover` button. A future update may make it possible to link this account to a Google / email account.


Geno22 (approved):
Yeah good we will implement later the anonymous house

PR #97, opened 2022-03-24
Feature/verif email

Defibrilator:
When registering with an email, it now sends a verification email and blocks the user from using the app without verifying his email.


app/src/main/java/com/github/houseorganizer/houseorganizer/MainActivity.java
```diff
@@ -6,27 +6,30 @@
...
         FirebaseUser user = mAuth.getCurrentUser();
 
         /* If user is not authenticated, send him to LoginActivity to authenticate first.
-        * Else send him to MainScreenActivity */
+         * Else send him to MainScreenActivity */
         (new Handler()).postDelayed(() -> {
             if (user != null) {
-                Intent mainScreenIntent = new Intent(this, MainScreenActivity.class);
-                startActivity(mainScreenIntent);
+                if (user.isEmailVerified()) {
```
aindreias, line 28: Maybe add one more check here: `if (user.isEmailVerified() || user.isAnonymous()))` s.t. anon users aren't redirected to LoginEmail

aindreias (approved):
Overall good, well-modularized code. Thank you for the contribution!

app/src/main/java/com/github/houseorganizer/houseorganizer/login/LoginEmail.java
```diff
@@ -33,10 +37,8 @@ protected void onCreate(Bundle savedInstanceState) {
     }
 
     private boolean inputsNotEmpty() {
-        EditText email_field = findViewById(R.id.log_enter_email);
-        EditText password_field = findViewById(R.id.log_enter_password);
-        String email = email_field.getText().toString();
-        String password = password_field.getText().toString();
+        String email = ((EditText) findViewById(R.id.log_enter_email)).getText().toString();
+        String password = ((EditText) findViewById(R.id.log_enter_password)).getText().toString();
```
Oxydeas, line 41: not sure you need casts here

Oxydeas (approved):
Good, nothing much to say as the implementation is pretty straight forward

app/src/main/java/com/github/houseorganizer/houseorganizer/login/LoginEmail.java
aindreias: I think they're necessary because `View` doesn't have a `getText` method

app/src/main/java/com/github/houseorganizer/houseorganizer/MainActivity.java
Defibrilator: good thinking i added it because its required but for now it makes the app crash because anonymous are not in houses

app/src/main/java/com/github/houseorganizer/houseorganizer/login/LoginEmail.java
Defibrilator: i do, otherwise they are typed as View

app/src/main/java/com/github/houseorganizer/houseorganizer/MainActivity.java
aindreias: Ok, let's mark it as a bug to fix in one of the upcoming sprints

PR #111, opened 2022-03-24
Feature/settings/link nickname to firebase

Oxydeas:
### Add the possibility for a user to define its nickname within the app
The user can decide of a nickname and it is pushed to firebase automatically
**NOT TESTED** Need firebase mock


aindreias (approved):
Thank you for the feature! I think this might not work with anonymous accounts since they don't have a linked email, but we can look into it next week.

Oxydeas:
> 

I don't know to which intent we'll use anonymous accounts but we might need to look into it yes

app/src/main/java/com/github/houseorganizer/houseorganizer/SettingsActivity.java
```diff
@@ -1,17 +1,29 @@
...
 
 import androidx.appcompat.app.ActionBar;
 import androidx.appcompat.app.AppCompatActivity;
+import androidx.preference.Preference;
 import androidx.preference.PreferenceFragmentCompat;
+import androidx.preference.PreferenceManager;
+
+import com.google.firebase.auth.FirebaseAuth;
+import com.google.firebase.auth.FirebaseUser;
+import com.google.firebase.firestore.DocumentReference;
```
Arkadyus, line 14: there's 3 unused imports here, would be nice to clean them now

app/src/main/java/com/github/houseorganizer/houseorganizer/SettingsActivity.java
```diff
@@ -24,10 +36,42 @@ protected void onCreate(Bundle savedInstanceState) {
         }
     }
 
-    public static class SettingsFragment extends PreferenceFragmentCompat {
+    public static class SettingsFragment extends PreferenceFragmentCompat implements SharedPreferences.OnSharedPreferenceChangeListener {
+
+        private FirebaseFirestore db;
+        private FirebaseUser Uid;
```
Arkadyus, line 42: isnt it simpler to store the mail here ? or even not save anything in a variable since the uid/mail is used once in this class

Arkadyus (changes requested):
Just a few imports to clean up and a variable im unsure of how to change but the rest seems good !

app/src/main/java/com/github/houseorganizer/houseorganizer/SettingsActivity.java
Oxydeas: Yes, done

app/src/main/java/com/github/houseorganizer/houseorganizer/SettingsActivity.java
Oxydeas: Fixed

Arkadyus (approved):
Thanks for the changes !

PR #112, opened 2022-03-24
Cleanup

Oxydeas:
Fixed the rest of the code to use emails as a default when referring to a user in firebase.


Defibrilator (approved):
Good change.

Arkadyus (approved):
Everything looks good, we'll have to care for anonymous users who dont have an email later on though

PR #113, opened 2022-03-24
Feature/info button

Oxydeas:
**Basic retrieval of info from current house**
Currently in a separate activity and unformatted.

### DIFF
--- a/app/src/main/res/layout/activity_info.xml
+++ b/app/src/main/res/layout/activity_info.xml
@@ -10,7 +10,7 @@
         android:id="@+id/info_text_view"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:text="No info yet!"
+        android:text="@string/no_household_info"
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent"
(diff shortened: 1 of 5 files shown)


sansive (commented):
Maybe it is best if you use the current string HOUSEHOLD to send the name of the house to the info activity and then retrieve the data

app/src/main/res/layout/activity_info.xml
```diff
@@ -0,0 +1,18 @@
...
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    tools:context=".InfoActivity">
+
+    <TextView
+        android:id="@+id/info_text_view"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:text="No info yet!"
```
aindreias, line 13: Maybe extract the "No info yet!" string

aindreias (approved):
Thanks for the contribution, it's short but useful.

Defibrilator (approved):
Good, nothing much to say, an info layout is a future task.

Oxydeas:
> Maybe it is best if you use the current string HOUSEHOLD to send the name of the house to the info activity and then retrieve the data

Yes, I probably will in a future task but for now I think it is fine. If I wanted to do that I would have to retrieve the database again in the info activity etc. which might not be ideal

app/src/main/res/layout/activity_info.xml
Oxydeas: Commit coming up - Thanks

PR #119, opened 2022-03-29
Fix RegisterEmail not registering

Defibrilator:
Email registering was not functioning because of a typo.


sansive (dismissed):

sansive (commented):
I still don't receive any verification email

sansive (approved):
Verification email was on the spam folder.

Oxydeas (approved):
I also tried earlier and this works. Thank you for the fix.

PR #125, opened 2022-03-30
Feature/camera

Arkadyus:
This branch adds the ability to add an image to firebase with the id from the event it is linked to as its name.
For now we can only link one image per event and cannot retrieve them.
The way the event id is sent back to the MainScreenActivity is a little hacky but an intent sounds overkill and potentially unefficient when we just are coming back from a popup dialog.

### DIFF
--- a/app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
+++ b/app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
@@ -5,13 +5,14 @@ import android.app.AlertDialog;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Bundle;
-import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.widget.EditText;
 import android.widget.TextView;
 import android.widget.Toast;
 
+import androidx.activity.result.ActivityResultLauncher;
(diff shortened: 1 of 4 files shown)


aindreias (approved):
Overall great starting point for the attachments feature, thank you for the contribution Thibaud!

Defibrilator (approved):
Effectively feels hacky haha :) but looks good !

app/src/main/java/com/github/houseorganizer/houseorganizer/MainScreenActivity.java
```diff
@@ -82,6 +85,18 @@ protected void onCreate(Bundle savedInstanceState) {
...
     }
 
+    private ActivityResultLauncher<String> registerForEventImage() {
+        // Prepare the activity to retrieve an image from the gallery
+        return registerForActivityResult(new ActivityResultContracts.GetContent(),
+                uri -> {
+                    // Store the image on firebase storage
+                    FirebaseStorage storage = FirebaseStorage.getInstance();
+                    // this creates the reference to the picture
+                    StorageReference imageRef = storage.getReference().child(calendarAdapter.eventToAttach + ".jpg");
```
Oxydeas, line 95: Are images from the gallery all jpgs ?

Oxydeas (approved):
Thank you for starting on this requirement of the project 👍

PR #146, opened 2022-03-31
Cleanup

Arkadyus:
Cleaned up the imports
Added better refreshing of the calendar
Added easier event editing

This work is not linked to a PR, it is work I did to make the demo look cleaner and it definitely makes sense to add it on main


app/src/main/java/com/github/houseorganizer/houseorganizer/EventsAdapter.java
```diff
@@ -111,26 +112,43 @@ private void eventButtonListener(Event event, View v) {
...
+                            .setPositiveButton(R.string.confirm, (editForm, editFormId) -> editEventAndDismiss(event, editForm, dialogView, position))
                             .setNegativeButton(R.string.cancel, (editForm, editFormId) -> dialog.dismiss())
                             .show();
                 }).show();
     }
 
-    private void editEventAndDismiss(String eventId, DialogInterface editForm, View dialogView, FirebaseFirestore db) {
+    private View createEditDialog(View v, Event event) {
+        LayoutInflater inflater = LayoutInflater.from(v.getContext());
+        View retView = inflater.inflate(R.layout.event_creation, null);
```
sansive, line 134: There is still a warning here about passing "null" as the view root

sansive (commented):
There are still some warnings in xml files too

Geno22 (approved):
Thanks the project really needed that !

app/src/main/java/com/github/houseorganizer/houseorganizer/EventsAdapter.java
Arkadyus: Fixing this in next commit

sansive (approved):
Nice work!

PR #149, opened 2022-03-31
Starting firebase emulator

Geno22:
Created utility functions for the Firebase Emulators and wrote a few tests for the MAinScreenActivity


app/src/androidTest/java/com/github/houseorganizer/houseorganizer/FirebaseTestsHelper.java
```diff
@@ -0,0 +1,89 @@
...
+    private static boolean databaseEmulatorActivated = false;
+
+    public static String TEST_USER_MAIL = "test@house.com";
+    public static String TEST_USER_PASSWORD = "Householder";
+    public static String TEST_HOUSEHOLD_NAME = "MyFavoriteHouse";
+
+    public static void startAuthEmulator(){
+        if(authEmulatorActivated) return;
+        FirebaseAuth.getInstance().useEmulator("10.0.2.2", 9099);
+    }
```
aindreias, line 37: Should this also have `authEmulatorActivated = true;` ?

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/FirebaseTestsHelper.java
```diff
@@ -0,0 +1,89 @@
...
+
+        Task<AuthResult> t = FirebaseAuth.getInstance().signInWithEmailAndPassword(TEST_USER_MAIL, TEST_USER_PASSWORD)
+                .addOnCompleteListener(task -> {
+                    if(task.getResult().getUser()==null){
+                        FirebaseAuth.getInstance().createUserWithEmailAndPassword(TEST_USER_MAIL, TEST_USER_PASSWORD);
+                    }else{
+                        FirebaseAuth.getInstance().signOut();
+                    }
+                });
+        Tasks.await(t);
```
Oxydeas, line 61: Are we sure the await terminates or time outs ?

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/MainActivityTest.java
```diff
@@ -3,44 +3,87 @@
...
+
+    /*@Before
+    public void initIntents(){
+        Intents.init();
+    }/*
+
     /* See logo */
     @Test
-    public void seeHouseLogo() {
+    public void test1seeHouseLogo() {
```
aindreias, line 53: Perhaps the test names in this file could be changed, but this is a task for another sprint.

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/MainScreenActivityTest.java
```diff
@@ -225,12 +255,12 @@ public void signOutButtonIsClickable(){
     }
 
     @Test
-    public void signOutButtonFiresRightIntent(){
+    public void zSignOutButtonFiresRightIntent(){
```
Oxydeas, line 258: This seems a bit weird having to start test names with a z in order to have it run after the other ones. If we go this way, we should at least normalize formatting i.e test01<name>; test02<name> etc.

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/FirebaseTestsHelper.java
Geno22: If it doesn't terminates the test will be written as failed at some point so I think this is not an issue. On the other hand if it is prefered I can use the awaits(task, timeout) surcharge so we are sure it does terminates.

aindreias (approved):
Thank you for your work, and especially for the helper methods! I hope they will come in handy for the next sprints.

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/MainScreenActivityTest.java
aindreias: Maybe there is a way to force the order of tests. We could investigate this in a future sprint

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/MainActivityTest.java
Geno22: Yes this whole file needs to be remade as hours of testing couldn't get it to work. The names were a (failed) attempt to order the execution.

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/MainScreenActivityTest.java
Geno22: Yes good idea. An alternative would be to use @Order(int) before each test but that would mean adding it to all the existing tests

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/FirebaseTestsHelper.java
Oxydeas: I think it would be better actually yes

PR #153, opened 2022-04-04
Refactor

Defibrilator:
#### Refactor of the java package. 

Separated the different classes into packages : 
- activity
- calendar
- house
- login
- shop
- task
- user
- util

Moved a couple functions but not much change since current classes are not very linked between different packages. Main changes were for the different kinds of messages.

### DIFF
--- a/app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
+++ b/app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
@@ -5,14 +5,10 @@ import static com.github.houseorganizer.houseorganizer.util.Util.getSharedPrefsE
 import static com.github.houseorganizer.houseorganizer.util.Util.logAndToast;
 
 import android.annotation.SuppressLint;
-import android.app.AlertDialog;
-import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.os.Bundle;
-import android.view.LayoutInflater;
 import android.view.View;
-import android.widget.EditText;
 
 import androidx.activity.result.ActivityResultLauncher;
 import androidx.activity.result.contract.ActivityResultContracts;
@@ -23,7 +19,6 @@ import androidx.recyclerview.widget.RecyclerView;
 
 import com.github.houseorganizer.houseorganizer.R;
 import com.github.houseorganizer.houseorganizer.calendar.Calendar;
-import com.github.houseorganizer.houseorganizer.calendar.Calendar.Event;
 import com.github.houseorganizer.houseorganizer.calendar.EventsAdapter;
 import com.github.houseorganizer.houseorganizer.house.HouseSelectionActivity;
 import com.github.houseorganizer.houseorganizer.shop.ShopItem;
@@ -34,7 +29,6 @@ import com.github.houseorganizer.houseorganizer.task.TaskListAdapter;
 import com.github.houseorganizer.houseorganizer.task.TaskView;
 import com.github.houseorganizer.houseorganizer.user.DummyUser;
 import com.github.houseorganizer.houseorganizer.user.User;
-import com.github.houseorganizer.houseorganizer.util.Util;
 import com.google.firebase.auth.FirebaseAuth;
 import com.google.firebase.auth.FirebaseUser;
 import com.google.firebase.firestore.DocumentReference;
@@ -45,8 +39,6 @@ import com.google.firebase.storage.StorageReference;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.HashMap;
-import java.util.Map;
 import java.util.Objects;
 
 @SuppressWarnings("unused")
@@ -186,35 +178,8 @@ public class MainScreenActivity extends AppCompatActivity {
     }
 
     private void addEvent(View v) {
-        LayoutInflater inflater = LayoutInflater.from(MainScreenActivity.this);
-        final View dialogView = inflater.inflate(R.layout.event_creation, null);
-        new AlertDialog.Builder(MainScreenActivity.this)
-                .setTitle(R.string.event_creation_title)
-                .setView(dialogView)
-                .setPositiveButton(R.string.add, (dialog, id) -> pushEventAndDismiss(dialog, dialogView, v))
-                .setNegativeButton(R.string.cancel, (dialog, id) -> dialog.dismiss())
-                .show();
-    }
-
-    private void pushEventAndDismiss(DialogInterface dialog, View dialogView, View v) {
-        Map<String, Object> data = new HashMap<>();
-        final String title = ((EditText) dialogView.findViewById(R.id.new_event_title)).getText().toString();
-        final String desc = ((EditText) dialogView.findViewById(R.id.new_event_desc)).getText().toString();
-        final String date = ((EditText) dialogView.findViewById(R.id.new_event_date)).getText().toString();
-        final String duration = ((EditText) dialogView.findViewById(R.id.new_event_duration)).getText().toString();
-        Map<String, String> event = new HashMap<>();
-        event.put("title", title);
-        event.put("desc", desc);
-        event.put("date", date);
-        event.put("duration", duration);
-        if (Util.putEventStringsInData(event, data)) {
-            dialog.dismiss();
-            return;
-        }
-        data.put("household", currentHouse);
-        db.collection("events").add(data)
-                .addOnSuccessListener(documentReference -> refreshCalendar(findViewById(R.id.refresh_calendar)));
-        dialog.dismiss();
+        calendar.addEvent(v, MainScreenActivity.this, db, currentHouse, calendarAdapter, Arrays.asList("MainScreenActivity",
+                "addEvent:failure"));
     }
 
     @SuppressWarnings("unused")
--- a/app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
+++ b/app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
@@ -43,8 +43,6 @@ import com.google.firebase.firestore.QueryDocumentSnapshot;
 import com.google.firebase.storage.FirebaseStorage;
 import com.google.firebase.storage.StorageReference;
 
-import java.time.LocalDateTime;
-import java.time.ZoneOffset;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -107,7 +105,7 @@ public class MainScreenActivity extends AppCompatActivity {
     @Override
     protected void onResume() {
         super.onResume();
-        refreshCalendar(findViewById(R.id.calendar));
+        refreshCalendar(findViewById(R.id.refresh_calendar));
     }
 
     private ActivityResultLauncher<String> registerForEventImage() {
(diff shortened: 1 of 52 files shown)


aindreias (approved):
This was a much needed refactoring. The changes are intuitive and clean - thank you very much for your work!

app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
```diff
@@ -221,9 +235,10 @@ public void refreshCalendar(View v) {
                         }
                         calendarAdapter.notifyDataSetChanged();
                         calendar.setEvents(newEvents);
-                    }
-                    else {
-                        Toast.makeText(v.getContext(), v.getContext().getString(R.string.refresh_calendar_fail), Toast.LENGTH_SHORT).show();
+                    } else {
+                        logAndToast(Arrays.asList("MainScreenActivity",
+                                "refreshCalendar:failureToRefresh"), task.getException(),
+                                v.getContext(), v.getContext().getString(R.string.refresh_calendar_fail));
```
Arkadyus, line 241: Can these methods not be moved in their respective packages ? They dont feel like they belong in MainScreenActivity to me

app/src/main/java/com/github/houseorganizer/houseorganizer/util/Util.java
```diff
@@ -20,4 +34,49 @@ public static boolean putEventStringsInData(Map<String, String> event, Map<Strin
...
+        EMAIL_USED,
+        INPUTS_EMPTY
+    }
+
+    public static void logAndToast(List<String> logTagAndMsg, Exception e, Context cx, String toastMsg) {
+        assert logTagAndMsg.size() == 2;
+
+        Log.w(logTagAndMsg.get(0), logTagAndMsg.get(1), e);
+        Toast.makeText(cx, toastMsg, Toast.LENGTH_SHORT).show();
+    }
```
Arkadyus, line 65: We can probably use the same message for the log and the toast, also I dont think putting the log tag and log message in a list makes sense other than for making codeclimate happy, we should talk about that and maybe ignore codeclimate
(or using the same message would also fix it)

app/src/main/java/com/github/houseorganizer/houseorganizer/util/Util.java
```diff
@@ -20,4 +34,49 @@ public static boolean putEventStringsInData(Map<String, String> event, Map<Strin
...
+                error_field.setText(R.string.email_not_valid);
+                break;
+            case INVALID_PASSWORD:
+                error_field.setText(R.string.password_not_valid);
+                break;
+            case EMAIL_USED:
+                error_field.setText(R.string.email_already_used);
+                break;
+            case INPUTS_EMPTY:
+                error_field.setText(R.string.inputs_empty);
```
Arkadyus, line 79: Im also not too sure about extracting this method since its one line and the line is different in every case

Arkadyus (changes requested):
Good refactoring, though i have a few questions (as comments) about it. Also please optimize the imports before merging
Also shouldnt we separate the xml files in packages as well ?

app/src/main/java/com/github/houseorganizer/houseorganizer/util/Util.java
```diff
@@ -20,4 +34,49 @@ public static boolean putEventStringsInData(Map<String, String> event, Map<Strin
         }
         return false;
     }
```
Arkadyus, line 36: This method might be able to be moved into the Event class also, since it is about events

app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
Defibrilator: Good idea

app/src/main/java/com/github/houseorganizer/houseorganizer/util/Util.java
Defibrilator: I don't think so, log is for the debugger and toast is for the user hence they cannot have the same message. 
We can talk about which type to use on our meeting.

app/src/main/java/com/github/houseorganizer/houseorganizer/util/Util.java
Defibrilator: Yeah I guess its not very useful but it does make the email login methods a tiny bit clearer. It was already extracted to LoginHelpers, I just moved it to Util to make it more generic.

app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
Defibrilator: But apparently Code Climate is not okay with that.

app/src/main/java/com/github/houseorganizer/houseorganizer/util/Util.java
Defibrilator: True

app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
aindreias: I had a similary problem last week, codeclimate only allows a max of 4 arguments per method. So it's difficult to move things around when you need a lot of outside information. Perhaps we could have some a controller or decorator for the calendar which receives many params from MainScreenActivity & uses those in its methods. Not sure if that's a good solution though

app/src/main/java/com/github/houseorganizer/houseorganizer/util/Util.java
Defibrilator: Okay I separated the list into two strings. I increased code climate limit on args because it made calendar methods fail.

app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
Defibrilator: For now I increased the limit to 7 but we can talk about it during our meeting.

app/src/main/java/com/github/houseorganizer/houseorganizer/calendar/Calendar.java
```diff
@@ -0,0 +1,228 @@
...
+                    } else {
+                        logAndToast(funcAndErrMessage.get(0), funcAndErrMessage.get(1), task.getException(),
+                                v.getContext(), v.getContext().getString(R.string.refresh_calendar_fail));
+                    }
+                });
+
+    }
+
+    public void addEvent(View v, Context ctx, FirebaseFirestore db, DocumentReference currentHouse,
+                         EventsAdapter calendarAdapter, List<String> funcAndErrMessage) {
```
Arkadyus, line 101: Same thing here for funcAndErrMessage

app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
```diff
@@ -168,65 +172,14 @@ private void rotateView(View v) {
...
-        new AlertDialog.Builder(MainScreenActivity.this)
-                .setTitle(R.string.event_creation_title)
-                .setView(dialogView)
-                .setPositiveButton(R.string.add, (dialog, id) -> pushEventAndDismiss(dialog, dialogView, v))
-                .setNegativeButton(R.string.cancel, (dialog, id) -> dialog.dismiss())
-                .show();
+    private void refreshCalendar(View v) {
+        calendar.refreshCalendar(findViewById(R.id.refresh_calendar), db, currentHouse, calendarAdapter, Arrays.asList("MainScreenActivity",
+                "refreshCalendar:failureToRefresh"));
     }
```
Arkadyus, line 178: Since the method is moved into Calendar, you can delete this one and replace the calls in the class by calls to calendar.refreshCalendar, the same applies for the other methods that were moved

app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
```diff
@@ -168,65 +172,14 @@ private void rotateView(View v) {
...
-                    }
-                    else {
-                        Toast.makeText(v.getContext(), v.getContext().getString(R.string.refresh_calendar_fail), Toast.LENGTH_SHORT).show();
-                    }
-                });
-
+    private void addEvent(View v) {
+        calendar.addEvent(v, MainScreenActivity.this, db, currentHouse, calendarAdapter, Arrays.asList("MainScreenActivity",
+                "addEvent:failure"));
     }
```
Arkadyus, line 183: Could the other methods not be moved for some reason or do you think they belong here ? Im hesitant with them but feel like they might be able to be moved (initializeTaskList, registerForEventImage, loadHousehold, rotateView and rotateLists)

app/src/main/java/com/github/houseorganizer/houseorganizer/calendar/Calendar.java
```diff
@@ -0,0 +1,228 @@
...
+    public String toString() {
+        StringBuilder ret = new StringBuilder("Calendar with view : " + view + " and the following events :\n");
+        for(Event event : events) {
+            ret.append(event).append("\n");
+        }
+        return ret.toString();
+    }
+
+    @SuppressLint("NotifyDataSetChanged")
+    public void refreshCalendar(View v, FirebaseFirestore db, DocumentReference currentHouse, EventsAdapter calendarAdapter, List<String> funcAndErrMessage) {
```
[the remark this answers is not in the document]
Defibrilator: We would need to increase the arg limit to 8, I dont think its worth it right now

app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
aindreias: I think the easier (or more difficult?) question is: what methods should be left in MainScreenActivity?

Defibrilator:
> Why is there an activity package but activities are split among other directories ?

Because activity is not the right name apparently, in this package we want the different panels inside the app so settings main screen, info, etc.. but for example every class inside login package is an activity.

app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
Defibrilator: My opinion is that initializeTaskList should be left in mainscreen, 
loadHousehold could be extracted but we would need to create a house class and since this method is only used here with a lot of internal variables I do not think it is beneficial.
I could move registerForEventImage to EventsAdapter but I'm not sure it would be beneficial, since you wrote the method here I'll leave it up to you to chose where and how it should be (in an other branch preferably).
I moved rotateView to the class Calendar.
I do not think rotateList should be moved since it manages both kinds of lists which are in different packages.

app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
Arkadyus: I think changing line 84 to a lambda makes more sense than keeping the method here personally

app/src/main/java/com/github/houseorganizer/houseorganizer/activity/MainScreenActivity.java
Arkadyus: This sounds good to me, ill eventually move registerForEventImage to EventsAdapter most likely (using the cleanup branch)

Oxydeas (approved):
Renamed the package and two other apporved. Good to me

Arkadyus (approved):
Looks good to me, if we ever decide on changing some methods' location it should be easier now
Thanks a lot for the refactoring

PR #155, opened 2022-04-05
Test/edit household test

Oxydeas:
400 lines of test which should cover everything in EditHousehold
Minor refactor of the helper to extract a method which I use before and after every test in the suite

### DIFF
--- a/app/src/androidTest/java/com/github/houseorganizer/houseorganizer/EditHouseholdTest.java
+++ b/app/src/androidTest/java/com/github/houseorganizer/houseorganizer/EditHouseholdTest.java
@@ -60,7 +60,8 @@ public class EditHouseholdTest {
     }
 
     @AfterClass
-    public static void signOut(){
+    public static void signOut() throws ExecutionException, InterruptedException {
+        FirebaseTestsHelper.createHouseholds();
         auth.signOut();
     }
 
@@ -69,11 +70,6 @@ public class EditHouseholdTest {
 
     @Before
     public void setupHouseholds() throws ExecutionException, InterruptedException {
-        FirebaseTestsHelper.createHouseholds();
-    }
-
-    @After
-    public void dismantleHouseholds() throws ExecutionException, InterruptedException {
         FirebaseTestsHelper.createHouseholds();
     }
 


app/src/androidTest/java/com/github/houseorganizer/houseorganizer/EditHouseholdTest.java
```diff
@@ -0,0 +1,389 @@
...
+        FirebaseTestsHelper.createHouseholds();
+    }
+
+    @After
+    public void dismantleHouseholds() throws ExecutionException, InterruptedException {
+        FirebaseTestsHelper.createHouseholds();
+    }
+
+    @Test
+    public void addUserIsDisplayed() {
```
aindreias, line 81: Do the display tests work all the time? I remember we had some problems with buttons on main screen failing this kind of test. Also, another question: what do you think of "squishing" displayed / clickable tests into one function? This is what I'm thinking of doing when testing the UI for task lists.

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/EditHouseholdTest.java
```diff
@@ -0,0 +1,389 @@
...
+        List<String> resident_after = (List<String>) houseData_after.get("residents");
+        Long num_residents_after = (Long) houseData_after.get("num_members");
+
+        // Compare states
+        assertEquals(resident_before, resident_after);
+        assertEquals(num_residents_before, num_residents_after);
+    }
+
+    @Test
+    public void removeUserFailsWhenRemovingYourself() throws ExecutionException, InterruptedException {
```
aindreias, line 293: Why is this the current behaviour? Is it because the only person able to remove household members is the owner => the owner can't remove him/herself??

aindreias (approved):
Thanks for the very detailed tests Luca! I have 2-3 questions in the comments, but only to help my comprehension of household editing

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/EditHouseholdTest.java
Oxydeas: Actually yes. I disallowed this because it would mean that we have to randomly assign someone else in the household as owner. What we can do in the future is a button to remove yourself from he household if you are a participant but no he owner i.e. you just quit the household

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/EditHouseholdTest.java
Oxydeas: Yes, the tests do work every time they are run I don’t need to use thread sleep for some reason

aindreias: I was just curious if you'd be open to it. In any case, I don't think there's a need to make these changes right now.

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/EditHouseholdTest.java
aindreias: Right, that would be a great feature! Imagine trying to quit a household but the owner won't kick you out :D

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/FirebaseTestsHelper.java
```diff
@@ -135,6 +138,28 @@ protected static void createTestTaskList() throws ExecutionException, Interrupte
...
+
+        createTestHouseholdOnFirestoreWithName(TEST_HOUSEHOLD_NAMES[1], TEST_USERS_EMAILS[0],
+                Arrays.asList(TEST_USERS_EMAILS[0], TEST_USERS_EMAILS[2]), TEST_HOUSEHOLD_NAMES[1]);
+
+        createTestHouseholdOnFirestoreWithName(TEST_HOUSEHOLD_NAMES[2], TEST_USERS_EMAILS[1],
+                Arrays.asList(TEST_USERS_EMAILS[1], TEST_USERS_EMAILS[2], TEST_USERS_EMAILS[3],
+                        TEST_USERS_EMAILS[4], TEST_USERS_EMAILS[5], TEST_USERS_EMAILS[6]),
+                TEST_HOUSEHOLD_NAMES[2]);
+    }
+
```
Geno22, line 156: Maybe it would be great to have an external way to know which user are in each house other than having to go inside the helper class
This could be done later though

Geno22 (approved):
Great thanks a lot for the tests and the helper !

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/EditHouseholdTest.java
```diff
@@ -0,0 +1,389 @@
...
+    public ActivityScenarioRule<EditHousehold> editHouseholdRule = new ActivityScenarioRule<>(intentFromHouseSelection);
+
+    @Before
+    public void setupHouseholds() throws ExecutionException, InterruptedException {
+        FirebaseTestsHelper.createHouseholds();
+    }
+
+    @After
+    public void dismantleHouseholds() throws ExecutionException, InterruptedException {
+        FirebaseTestsHelper.createHouseholds();
```
aindreias, line 77: Just caught this one - shouldn't you delete the data instead of creating it again?

aindreias (changes requested):
Updating my review since there seems to be a problem with the @After method

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/EditHouseholdTest.java
Geno22: I think the function just has a confusing name because actually it sets up the households to their original state.
It creates them only if they don't exist, otherwise it resets them

Oxydeas: Yes @Geno22 is right. The create household does not care about the state of the data base as the state is different after each test. It does not try to create, but rathers sets the data to X without caring about any state whatsoever

aindreias: OK, I see now. The thing is, you can add this to @AfterClass instead. Since the @Before already calls the same method for each test, an @After seems like a surplus to me in this context?

Oxydeas: On it

aindreias (approved):
All good for me now!

PR #158, opened 2022-04-06
Better UI of the info button using recyler views

Oxydeas:
Small feature that displays the residents on the info button screen.

### DIFF
--- a/app/src/main/res/layout/user_row.xml
+++ b/app/src/main/res/layout/user_row.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="wrap_content">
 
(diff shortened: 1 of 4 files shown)


Geno22 (approved):
Yeah great works well and looks better than before thanks !

app/src/main/res/values/strings.xml
```diff
@@ -11,7 +11,7 @@
     <string name="img_house_row_desc">Image for the given house name</string>
     <string name="home_logo">Home Logo</string>
     <string name="title_activity_settings">Settings</string>
-    <string name="no_household_info">No info yet!</string>
```
sansive, line 14: I would keep this string and use it as a default message when the user doesn't have a household selected, for example when he isn´t register in any

app/src/main/res/layout/user_row.xml
```diff
@@ -15,22 +16,19 @@
 
         <androidx.constraintlayout.widget.ConstraintLayout
             android:layout_width="match_parent"
-            android:layout_height="match_parent">
+            android:layout_height="match_parent"
+            android:layout_gravity="center">
 
             <TextView
```
aindreias, line 22: Could you use some sort of android:gravity="center_vertical" ? this way the members' emails would start from the left. This might be just a stylistic preference though

aindreias (approved):
Thanks for the change Luca! This will also be useful for viewing assignees, actually

app/src/main/res/values/strings.xml
Oxydeas: Since we're using a recycler view instead of a text view now, if there is no household selected then the members will just be empty

app/src/main/res/layout/user_row.xml
Oxydeas: Done

app/src/main/res/values/strings.xml
Oxydeas: I added it for good measure

sansive: Perfect!

sansive (approved):

app/src/main/res/layout/user_row.xml
aindreias: Thanks Luca!

PR #159, opened 2022-04-06
Local storage

Oxydeas:
This is related to my task "investigate local storage on the device". A new package was added "storage" where there are two classes. LocalStorage offers a helper to store text files on the device itself. LocalCache offers an interface to interact with a shared cache which is alive only during the lifetime of the app. The idea of the enum in the shared cache is to have a common ground for the keys we use, instead of just random strings.


aindreias (approved):
Nice prototype, thanks for the good work!

Arkadyus (approved):
Looks clean to me, good job !

PR #163, opened 2022-04-07
Bug/default house

sansive:
## Task
* Change current household to the recently added one when adding a household
* Implement default house when user sign in


Arkadyus (approved):
Looks good and it works well, thanks for the fix

aindreias (approved):
Great work, thanks for the fix!

PR #181, opened 2022-04-13
Features/calendar main view

Arkadyus:
Add a navigation bar at the bottom of the screen
Add a calendar fullscreen view that you can access through the bar
Edit the upcoming view for the calendar
Remove the refresh and the cycle view buttons in MainScreenActivity
Refactor a lot in EventsAdapter
Edit and add a few tests accordingly

!! Bump the LOC limit to from 250 to 350 !!
this is because after asking the group it has been agreed on to bump it as there does not seem to be an intuitive way of moving lines from EventsAdapter to another class


Geno22 (commented):
image
This does not work for now I will come back later

Geno22 (approved):
This is very good and clean thank you !

aindreias (approved):
Nice work, the tests are very thorough!

PR #182, opened 2022-04-14
Feature/link task lists with households and users

aindreias:
**Changes**
- task list data is now broken up into 1 document per task (/task_dump/...) and a metadata file (/task_lists/...), which stores among other things a list of task pointers
 - `HTask` no longer uses the `User` interface for owners & assignees, instead it uses a `String`
 - UI + DB reflection tests for the task list fragment on MainScreenActivity

**Bugs** _(not tested, to be addressed this/next sprint)_
- changing the title of a task crashes the app (due to a `Tasks::await` call in `FirestoreTask`)
- removing a task doesn't update the pointer list in the metadata file

### DIFF
--- a/app/src/androidTest/java/com/github/houseorganizer/houseorganizer/FirebaseTestsHelper.java
+++ b/app/src/androidTest/java/com/github/houseorganizer/houseorganizer/FirebaseTestsHelper.java
@@ -101,7 +101,7 @@ public class FirebaseTestsHelper {
     }
 
     private static void wipeTasksThenMetadata(DocumentSnapshot tlDocSnap) throws ExecutionException, InterruptedException {
-        List<String> taskPtrs = (List<String>)
+        List<DocumentReference> taskPtrs = (List<DocumentReference>)
                 Objects.requireNonNull(tlDocSnap.getData()).getOrDefault("task-ptrs", new ArrayList<>());
 
         CollectionReference taskDumpRef =
@@ -109,8 +109,8 @@ public class FirebaseTestsHelper {
                         .collection("task_dump");
 
         assert taskPtrs != null;
-        for (String taskPtr : taskPtrs) {
-            Tasks.await(taskDumpRef.document(taskPtr).delete());
+        for (DocumentReference taskPtr : taskPtrs) {
+            Tasks.await(taskPtr.delete());
         }
 
         Tasks.await(tlDocSnap.getReference().delete());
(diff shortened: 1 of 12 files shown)


app/src/androidTest/java/com/github/houseorganizer/houseorganizer/FirebaseTestsHelper.java
```diff
@@ -324,18 +306,9 @@ protected static void setUpFirebase() throws ExecutionException, InterruptedExce
...
-                Arrays.asList(TEST_USERS_EMAILS[0], TEST_USERS_EMAILS[2]), TEST_HOUSEHOLD_NAMES[1]);
+        createHouseholds();
 
-        createTestHouseholdOnFirestoreWithName(TEST_HOUSEHOLD_NAMES[2], TEST_USERS_EMAILS[1],
-                Arrays.asList(TEST_USERS_EMAILS[1], TEST_USERS_EMAILS[2], TEST_USERS_EMAILS[3],
-                        TEST_USERS_EMAILS[4], TEST_USERS_EMAILS[5], TEST_USERS_EMAILS[6]),
-                TEST_HOUSEHOLD_NAMES[2]);
-
-        createTestTaskList();
+        //createTestTaskList(); [MOVED TO `CREATE TEST HH` SINCE TLs NOW LINKED TO HHs]
```
Arkadyus, line 311: You can probably remove this line instead of commenting it

app/src/main/java/com/github/houseorganizer/houseorganizer/task/TaskAssigneeAdapter.java
```diff
@@ -44,7 +43,9 @@ public void onBindViewHolder(@NonNull BiViewHolder<TextView, ImageButton> holder
         boolean isAssigned = isAmongAssignees(assigneeEmail);
         assigneeEmailTextView.setText(assigneeEmail);
 
-        imgButton.setBackgroundResource(isAssigned ? R.drawable.remove_person : R.drawable.add_person);
+        int rscId = isAssigned ? R.drawable.remove_person : R.drawable.add_person;
+        imgButton.setBackgroundResource(rscId);
+        imgButton.setTag(rscId);
```
Arkadyus, line 48: it might look cleaner if you chain these calls like imgButton.setBackgroundResource(rscId).setTag(rscId);

Arkadyus (changes requested):
Looks good, the string resources R.string.no_household_info and R.string.subtask_completion_desc are unused so it would be good to remove them

app/src/main/java/com/github/houseorganizer/houseorganizer/task/TaskAssigneeAdapter.java
aindreias: Unfortunately setBackgroundResource has a void return :(

aindreias:
> Looks good, the string resources R.string.no_household_info and R.string.subtask_completion_desc are unused so it would be good to remove them

Removing R.string.no_household_info; R.string.subtask_completion_desc should be used & I'm making the changes in the next commit. Thanks a lot for the catch!

Arkadyus (approved):
looks clean now, thanks!

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/TaskListOnMainScreenTest.java
```diff
@@ -30,10 +38,15 @@
...
+
 @RunWith(AndroidJUnit4.class)
 public class TaskListOnMainScreenTest {
 
-    private static FirebaseFirestore db;
+    public static final String NEW_TASK_TITLE = "this is my task now";
+    public static final String NEW_SUBTASK_TITLE = "a new subtask!";
     private static FirebaseAuth auth;
 
     @BeforeClass
```
Oxydeas, line 52: Do you create a new tasklist on firebase in this test ? If so, is it deleted afterwards ?

Oxydeas (changes requested):
Looks good overall, we'll be able to delete the user class entirely after shopItems get revamped too!

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/TaskListOnMainScreenTest.java
aindreias: No new task list is created. All actions on the database by this test class are reverted in their respective methods.

Oxydeas: Oh I see you do it in each test, ok!

Oxydeas (approved):

PR #188, opened 2022-04-15
Feature/add dialog

sansive:
## Task
Implement add dialog that redirects to the create household screen when the user doesn't have any household.

## Testing
Done in `MainScreenActivityTest.java`


app/src/androidTest/java/com/github/houseorganizer/houseorganizer/FirebaseTestsHelper.java
```diff
@@ -202,6 +202,24 @@ protected static Map<String, Object> fetchHouseholdData(String houseName, Fireba
...
+        return task.getResult().getDocuments();
+    }
+
+    protected static boolean householdExists(String houseName, FirebaseFirestore db) throws ExecutionException, InterruptedException {
+        Task<DocumentSnapshot> task = db.collection("households").document(houseName).get();
+        Tasks.await(task);
+        return task.getResult().exists();
+    }
+
+    protected static boolean EventExists(String event, FirebaseFirestore db) throws ExecutionException, InterruptedException {
```
Geno22, line 217: Just for formating but the name of a function should not start with an uppercase letter

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/MainScreenActivityTest.java
```diff
@@ -45,6 +50,9 @@ public static void createMockFirebase() throws ExecutionException, InterruptedEx
 
         db = FirebaseFirestore.getInstance();
         auth = FirebaseAuth.getInstance();
+
+        auth.signOut();
+        FirebaseTestsHelper.signInTestUserWithCredentials(FirebaseTestsHelper.TEST_USERS_EMAILS[7], FirebaseTestsHelper.TEST_USERS_PWD[7]);
```
Geno22, line 55: I am afraid that signin-in with a particular user before beginning the tests might break them. Maybe only sign-in the user at the start of the tests that need it ?

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/MainScreenActivityTest.java
```diff
@@ -151,6 +159,17 @@ public void cycleIsDisplayed() {
...
 
+    @Test
+    public void addDialog() throws ExecutionException, InterruptedException {
+        Intents.init();
+        onView(withText("Add household")).perform(click());
+        intended(hasComponent(CreateHouseholdActivity.class.getName()));
+        Intents.release();
+
+        auth.signOut();
+        FirebaseTestsHelper.signInTestUserWithCredentials(FirebaseTestsHelper.TEST_USERS_EMAILS[0], FirebaseTestsHelper.TEST_USERS_PWD[0]);
```
Geno22, line 170: Same here after this test the connected user is not the same as at the beginning. But as the tests are done in a random order we cannot know which user is connected

app/src/main/java/com/github/houseorganizer/houseorganizer/house/EditHousehold.java
```diff
@@ -171,34 +165,99 @@ public void removeUser(View view) {
...
+                .whereEqualTo("household", currentHousehold)
+                .get().addOnCompleteListener(task1 -> {
+            if (task1.isSuccessful()) {
+                for (QueryDocumentSnapshot document : task1.getResult()) {
+                    firestore.collection("events").document(document.getId()).delete()
+                            .addOnCompleteListener(task2 -> {
+                                if(!task2.isSuccessful()) {
+                                    Toast.makeText(getApplicationContext(), view.getContext().getString(R.string.remove_calendar_failure), Toast.LENGTH_SHORT).show();
+                                }
+                            });
```
Geno22, line 238: [OPTIONAL] Instead of using two onCompleteListeners you can use continueWithTask, return a taask and then put an onComplete listener on everything
I spent quite som etime figuring out how it works so if you want you can check MainScreenActivity in the branch groceries_refactor

Geno22 (changes requested):
Thanks a lot good job ! The few changes aren't much I hope for you that the tests will pass ^^

Geno22 (approved):
Perfect thanks !

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/NoHouseholdsCaseTest.java
```diff
@@ -0,0 +1,69 @@
...
+public class NoHouseholdsCaseTest {
+
+    private static FirebaseFirestore db;
+    private static FirebaseAuth auth;
+
+    @BeforeClass
+    public static void createMockFirebase() throws ExecutionException, InterruptedException {
+        FirebaseTestsHelper.startAuthEmulator();
+        FirebaseTestsHelper.startFirestoreEmulator();
+        FirebaseTestsHelper.setUpFirebase_noHouseholds();
```
Oxydeas, line 46: Replace by the usual setup firebase but use a user who has no household associated

Oxydeas (changes requested):

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/NoHouseholdsCaseTest.java
Oxydeas: setUpFirebase
signOut
signIn as user who has no household

Oxydeas (approved):
Thank you for fixing!

PR #193, opened 2022-04-16
Feature/groceries refactor

Geno22:
Refactor of groceries on Firebase and added synchronized list to the MainScreen.
Not tested yet


app/src/androidTest/java/com/github/houseorganizer/houseorganizer/FirestoreShopListTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @BeforeClass
+    public static void startingFirebase() throws ExecutionException, InterruptedException {
+        FirebaseTestsHelper.startAuthEmulator();
+        FirebaseTestsHelper.startFirestoreEmulator();
+        FirebaseTestsHelper.setUpFirebase();
+        db = FirebaseFirestore.getInstance();
+        Task<QuerySnapshot> t = db.collection("shop_lists").get();
+        Tasks.await(t);
+        assertThat(t.getResult().getDocuments().size(), is(1));
```
sansive, line 40: Asserting things in this type of methods can make all the test fails, did you have that in mind while doing it?

app/src/main/java/com/github/houseorganizer/houseorganizer/shop/FirestoreShopList.java
```diff
@@ -0,0 +1,154 @@
...
+            items.add(itemMap);
+        }
+        return items;
+    }
+
+    private static List<ShopItem> convertFirebaseListToItems(List<Map<String, Object>> list){
+        List<ShopItem> items = new LinkedList<>();
+        for(Map<String, Object> m : list){
+            items.add(new ShopItem((String)m.get("name"),
+                    new Long((long) m.get("quantity")).intValue(),
```
sansive, line 115: Here Cirrus suggests to use `Long.valueOf((long) m.get("quantity"))` instead

app/src/main/java/com/github/houseorganizer/houseorganizer/shop/ShopList.java
```diff
@@ -37,23 +31,24 @@ public void removeItem(int index){
...
-    //Getters
-    public String getListName() {
-        return name;
+    public List<ShopItem> getItems(){
+        return new ArrayList<>(items);
     }
 
-    public User getOwner() {
-        return owner;
+    public void setItems(List<ShopItem> items) {
```
sansive, line 46: This method sets the entire list of items, what if I only want to change one of them. Do we have a case for that?

sansive (commented):
Overall great work! I just had some questions. Also, there are several warnings about autofillHints and accessibility labels on the xml files

app/src/main/java/com/github/houseorganizer/houseorganizer/shop/ShopList.java
Geno22: Yes if you want to change just one of them you can do
 `shopList.getItem(index).setName("")` for example and then
 `shopList.updateItems` to upload the changes to Firebase ^^

app/src/main/java/com/github/houseorganizer/houseorganizer/shop/FirestoreShopList.java
Geno22: This will give a long instead of a Long which implies I would have to recast it in order to have the IntegerValue ^^'

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/FirestoreShopListTest.java
Geno22: Nice catch I will make the test less specific but if there are 0 shopLists it is normal that everything fails as I cannot compare the local data to the remote one.

sansive (approved):

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/MainActivity.java
```diff
@@ -26,11 +26,12 @@ protected void onCreate(Bundle savedInstanceState) {
...
             if (user != null) {
-                if (user.isEmailVerified() || user.isAnonymous()) {
+                // TODO: I can't test with this can we remove it for now while keeping the TODO ?
+                /*if (user.isEmailVerified() || user.isAnonymous()) {*/
                     startActivity(new Intent(MainActivity.this, MainScreenActivity.class));
-                } else {
+                /*} else {
                     startActivity(new Intent(MainActivity.this, LoginEmail.class));
-                }
+                }*/
```
Arkadyus, line 34: I think its better to keep this one uncommented and to not test for now as commenting this makes our app less usable and more buggy potentially

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/MainScreenActivity.java
```diff
@@ -83,11 +92,43 @@ protected void onCreate(Bundle savedInstanceState) {
         findViewById(R.id.calendar_view_change).setOnClickListener(v -> calendar.rotateCalendarView(v, this, calendarAdapter, calendarEvents));
         findViewById(R.id.add_event).setOnClickListener(this::addEvent);
         findViewById(R.id.refresh_calendar).setOnClickListener(this::refreshCalendar);
-        findViewById(R.id.new_task).setOnClickListener(v -> TaskView.addTask(db, taskList, taskListAdapter, listView));
+        //findViewById(R.id.new_task).setOnClickListener(v -> TaskView.addTask(db, taskList, taskListAdapter, listView));
```
Arkadyus, line 95: why is this commented out ?

app/src/main/java/com/github/houseorganizer/houseorganizer/shop/FirestoreShopList.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    public static FirestoreShopList buildShopList(DocumentSnapshot documentSnapshot){
+        if(documentSnapshot == null) return null;
+        DocumentReference household = (DocumentReference) documentSnapshot.get("household");
+        List<Map<String, Object>> list = (List<Map<String, Object>>) documentSnapshot.get("items");
+        List<ShopItem> items = convertFirebaseListToItems(list);
+        return new FirestoreShopList(household, documentSnapshot.getReference(), items);
+    }
+}
```
Arkadyus, line 154: Is it necessary to have both a FireStore shoplist and a non FireStore one ? I feel like only one would suffice

app/src/main/java/com/github/houseorganizer/houseorganizer/shop/ShopListView.java
```diff
@@ -0,0 +1,48 @@
...
+        try {
+            quantity = Integer.parseInt(((EditText) dialogView.findViewById(R.id.editTextQuantity)).getText().toString());
+        }catch (Exception e){
+            // Only possible bad input is empty field
+        }
+        shopList.addItem(new ShopItem(name, quantity, unit));
+        adapter.notifyItemInserted(shopList.size()-1);
+        shopList.updateItems();
+    }
+}
```
Arkadyus, line 48: These methods might make more sense in ShopListAdapter as they all have the adapter as an argument, I might be missing a reason though of course

app/src/main/res/layout/shoplist_row.xml
```diff
@@ -1,14 +1,26 @@
...
+    android:id="@+id/shopDeleteItem"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
     android:paddingTop="0dp"
     android:paddingBottom="0dp">
 
     <CheckBox
         android:id="@+id/item_checkbox"
-        android:layout_width="0dp"
+        android:layout_width="308dp"
```
Arkadyus, line 12: using relative sizes might be better here as absolute ones wont change depending on the size of the phone

Arkadyus (changes requested):
Nice edit, put quite a few comments though as I dont understand/am confused about some parts of the code

app/src/main/java/com/github/houseorganizer/houseorganizer/shop/FirestoreShopList.java
aindreias: I think he wanted to keep the model separate from the implementation-specific use of Firebase

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/MainActivity.java
Geno22: I don't understand when this is un commentés i Just can't use thé app. Once i start it it takes me to the loginEmail class but si ce i log in with Google I can't login with e-mail and password. What I removed was Just the part that check if thé email is vérifié which we can't do right now

app/src/main/java/com/github/houseorganizer/houseorganizer/shop/FirestoreShopList.java
Geno22: Yes exactly but I admit that the New version of shoplist is very close to à simple list which was not the case before

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/MainScreenActivity.java
Geno22: This line should be deleted Now there is the function bottomAddButton that does this

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/MainActivity.java
Arkadyus: does it still happen after wiping out the data of the emulator ? We need to make the back button go to the login activity and not the main activity in that case though thats something to do for another time.
If the issue is still there after wiping the data tell me what to do to reproduce the issue and i can try to find a fix but im not sure we can let every non email verified or non anonymous user get on the mainscreen safely

Geno22: It is not for testing on the emulator but with live Firebase
Like when i start the app myself it takes me to the loginEmail which i can't use.
To avoid that I have to wipe the app data everytime but this is costly when I just want to see if one button works.
I personnally believe that we should either find a way to verify our emails or disable this while the app is not available to gain a ton of time during manual testing. 
But I guess if I am the only one complaining it must be that I am missing something. How do you do manual/visual tests ? doesn't it redirect you to loginEmail ?

app/src/main/res/layout/shoplist_row.xml
Geno22: I tried but could not get it to work. Maybe you could show me next week ?

app/src/main/java/com/github/houseorganizer/houseorganizer/shop/ShopListView.java
Geno22: Done thanks !

app/src/main/res/layout/shoplist_row.xml
Arkadyus: i just noticed this is in a linearLayout which makes it different than the ones i did, we can try and revamp that in another PR

Geno22: I finally found out how to do it so it should be good now ^^

Arkadyus (approved):
perfect, thanks for the changes !

PR #201, opened 2022-04-24
No household test case

sansive:
Add test for the case when the user doesn't have any household


Geno22 (approved):
Thanks for making it work !

Arkadyus (approved):
it feels weird to have a file just for this but it is necessary so thanks on the addition !

PR #205, opened 2022-04-26
Feature/groceries refactor

Geno22:
Added new screen for the groceries plus tested and refactored the groceries lists


app/src/androidTest/java/com/github/houseorganizer/houseorganizer/TaskListOnMainScreenTest.java
```diff
@@ -56,6 +56,7 @@ public static void createMockFirebase() throws ExecutionException, InterruptedEx
         FirebaseTestsHelper.setUpFirebase();
 
         auth = FirebaseAuth.getInstance();
+        FirestoreTaskTest.createMockFirebase();
```
aindreias, line 59: I don't know if this call is really necessary (line 59). How did it help when adding it?

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/GroceriesActivityTest.java
```diff
@@ -0,0 +1,136 @@
...
+        onView(withId(R.id.groceries_recycler)).check(matches(hasDescendant(withText(containsString("kg")))));
+
+        onView(withId(R.id.groceries_recycler))
+                .perform(RecyclerViewActions.actionOnItemAtPosition(
+                        1,
+                        RecyclerViewHelperActions.clickChildViewWithId(R.id.delete_item_button)));
+    }
+
+    @Test
+    public void navBarTakesBackToMainScreen(){
```
aindreias, line 107: Nice test case! I should add this to my TaskListActivity test as well

aindreias (changes requested):
(Approving once you reply to the comments) Everything seems good to me, and nice work on the tests!

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/TaskListOnMainScreenTest.java
Geno22: It helped because I kept having an error in this class that said that db was null. When adding the line it allowed db to be initialized thus not having the error anymore.
Maybe it was just a local thing but it fixed it on my machine ^^

Oxydeas (approved):
Good revamp, I don't have much to say as it is mostly the same code and testing 👍

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/GroceriesActivityTest.java
```diff
@@ -0,0 +1,136 @@
...
+        Intents.init();
+        onView(withId(R.id.nav_bar_menu)).perform(click());
+        intended(hasComponent(MainScreenActivity.class.getName()));
+        Intents.release();
+    }
```
Arkadyus, line 112: im not sure how important it is but if you test main screen you can also probably test the calendar screen

aindreias (approved):
Thanks for the explanation! Looks all good to me.

app/src/main/java/com/github/houseorganizer/houseorganizer/shop/GroceriesActivity.java
```diff
@@ -0,0 +1,75 @@
...
+                    if(t.isSuccessful()){
+                        shopList = t.getResult().getFirestoreShopList();
+                        shopListAdapter = t.getResult();
+                        shopList.getOnlineReference().addSnapshotListener((d, e) -> {
+                            shopList = FirestoreShopList.buildShopList(d);
+                            shopListAdapter.setShopList(shopList);
+                        });
+                        view.setLayoutManager(new LinearLayoutManager(this));
+                        view.setAdapter(shopListAdapter);
+                    }else Log.e("Groceries", "Could not initialize list");
```
Arkadyus, line 71: using logAndToast might be nice here, so the user knows that there was a bug when it cannot be initialized

app/src/main/res/layout/shop_item_dialog.xml
```diff
@@ -7,29 +7,29 @@
...
-        android:id="@+id/editTextName"
+        android:id="@+id/edit_text_name"
         android:layout_width="200dp"
         android:layout_height="wrap_content"
         android:layout_weight="1"
         android:ems="10"
-        android:inputType="textPersonName"
-        android:text="@string/name" />
+        android:hint="@string/name"
+        android:inputType="textPersonName" />
```
Arkadyus, line 16: Cirrus complains about missing autofillHints, its just adding a string that describes the content of the editText (you can probably put the same as for hint)
Since its really simple and doesnt cost anything to add we might as well add them
The same goes for the 2 other EditTexts further down

Arkadyus (changes requested):
I actually really like the way it looks and how smoothly it works, great job on that one.
Just added a few comments about small changes/questions

app/src/main/java/com/github/houseorganizer/houseorganizer/shop/GroceriesActivity.java
Geno22: Is logAndToast a function somewhere or should I just add a Toast after the log ?

app/src/main/res/layout/shop_item_dialog.xml
Geno22: Alright it is added

app/src/main/java/com/github/houseorganizer/houseorganizer/shop/GroceriesActivity.java
Geno22: Alright it is added thanks ^^

Arkadyus (approved):
everything looks good now, thanks!

PR #206, opened 2022-04-27
Features/calendar monthly view

Arkadyus:
This PR adds functionality to the monthly view : 
Days with events have a "!" on their button
When clicking on a day with events, it opens an upcoming view of the events on that day


aindreias (approved):
Thanks for the great feature, it must've been a pain to work with 3 different calendar adapters

app/src/main/java/com/github/houseorganizer/houseorganizer/calendar/CalendarAdapter.java
```diff
@@ -0,0 +1,125 @@
...
+                .setTitle(R.string.event_creation_title)
+                .setView(dialogView)
+                .setPositiveButton(R.string.add, (dialog, id) -> {
+                    Task<DocumentReference> pushTask = pushEventFromDialog(dialogView, currentHouse);
+                    if (pushTask != null) {
+                            pushTask.addOnSuccessListener(documentReference -> refreshCalendarView(ctx, currentHouse, errMessage, calendar.getView() == Calendar.CalendarView.MONTHLY));
+                    }
+                    dialog.dismiss();
+                })
+                .setNegativeButton(R.string.cancel, (dialog, id) -> dialog.dismiss())
```
sansive, line 66: I think it would also be nice if we do '.setCancelable(true);' so that if the user clicks in other parts of the screen the dialog disappears.

sansive (approved):
Great work! We should find an alternative for the thread.sleep

PR #208, opened 2022-04-27
Feature/info free text

Oxydeas:
Adds a free text field as attribute of a household. Displayed in the info screen and can be changed by any of the residents also in the info screen. Modifed tests and firebase helper to reflect this change on the emulators. Also tested the whole info activity now that we have the tools to test recycler views.

### DIFF
--- a/app/src/main/java/com/github/houseorganizer/houseorganizer/panels/InfoActivity.java
+++ b/app/src/main/java/com/github/houseorganizer/houseorganizer/panels/InfoActivity.java
@@ -2,6 +2,8 @@ package com.github.houseorganizer.houseorganizer.panels;
 
 import android.content.SharedPreferences;
 import android.os.Bundle;
+import android.view.View;
+import android.widget.EditText;
 import android.widget.TextView;
 
 import androidx.appcompat.app.AppCompatActivity;
@@ -15,7 +17,9 @@ import com.google.firebase.firestore.DocumentReference;
 import com.google.firebase.firestore.DocumentSnapshot;
 import com.google.firebase.firestore.FirebaseFirestore;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 public class InfoActivity extends AppCompatActivity {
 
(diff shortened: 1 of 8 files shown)


Oxydeas:
> Everything looking good to me, though it might make more sense to be able to edit the text by simply clicking on it, this can be done later when we try to make the app more user-friendly/revamp the UI

I thought about this but decided against it as it would imply listening to changes on the text field, so if a user clicks on the text box without paying attention and we register his changes, it could erase what was already written. I opted for a separate text box with confirm button.
But sure, this can be changed later if another alternative is judged more user friendly

aindreias:
> > Everything looking good to me, though it might make more sense to be able to edit the text by simply clicking on it, this can be done later when we try to make the app more user-friendly/revamp the UI
> 
> I thought about this but decided against it as it would imply listening to changes on the text field, so if a user clicks on the text box without paying attention and we register his changes, it could erase what was already written. I opted for a separate text box with confirm button. But sure, this can be changed later if another alternative is judged more user friendly

That's a very interesting perspective Luca, and you're right, for a house description we might want to be more cautious and prevent this type of mistake from happening.

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/InfoActivity.java
```diff
@@ -51,6 +61,19 @@ protected void onCreate(Bundle savedInstanceState) {
         }
     }
 
+    public void changeNotesHousehold(View view){
+        EditText editNotes = findViewById(R.id.editTextHouseholdNotes);
```
sansive, line 65: Why not use notesTextView? This will enhance the user experience for example in the case the user just wants to add something to the original notes.

Oxydeas: This would be impossible to implement without refactoring everything. Instead, When loading the notes I put the same text in the edit text box, which achieve the same "appending text" functionality

sansive: I understand

sansive (approved):

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/InfoActivity.java
Oxydeas: Took 1 hour to change the tests and everything but is done

PR #226, opened 2022-04-28
Feature/location

sansive:
## Task
* Implement use of GPS
* Add latitude and longitude to households


app/src/androidTest/java/com/github/houseorganizer/houseorganizer/FirebaseTestsHelper.java
```diff
@@ -211,9 +217,16 @@ protected static void createTestShopList() throws ExecutionException, Interrupte
         Task<DocumentReference> t = FirestoreShopList.storeNewShopList(db.collection("shop_lists"), shopList, household);
         Tasks.await(t);
         shopList.setOnlineReference(t.getResult());
+
+        household = db.collection("households").document(TEST_HOUSEHOLD_NAMES[1]);
+        shopList = new FirestoreShopList(household);
+        shopList.addItem(TEST_ITEM);
+        t = FirestoreShopList.storeNewShopList(db.collection("shop_lists"), shopList, household);
```
Geno22, line 224: Actually it is not necessary to add a shopList to a new house as it is added anyway (without item) if it does not exist by the MainScreen. However you might leave it like that so this  house already has an item

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/HouseSelectionActivityTest.java
```diff
@@ -36,6 +37,9 @@ public class HouseSelectionActivityTest {
     private static FirebaseFirestore db;
     private static FirebaseAuth auth;
 
+    @Rule
+    public GrantPermissionRule permissionRules = GrantPermissionRule.grant(android.Manifest.permission.ACCESS_FINE_LOCATION);
```
Geno22, line 41: So we have to put the rule of the premission in the tests class concerned ?

app/src/main/java/com/github/houseorganizer/houseorganizer/house/HouseSelectionActivity.java
```diff
@@ -32,21 +41,77 @@
...
+
+        requestPermission();
+        getCoordinates();
+        setHousesView();
+    }
+
+    private void requestPermission() {
+        if (ActivityCompat.checkSelfPermission(this, Manifest.permission.ACCESS_FINE_LOCATION) != PackageManager.PERMISSION_GRANTED) {
+            // Permissions not granted yet
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
Geno22, line 82: Cirrus says that this check is not necessary as "SDK_INT is always >= 26" but maybe there is something else ^^

app/src/main/java/com/github/houseorganizer/houseorganizer/house/HouseSelectionActivity.java
```diff
@@ -32,21 +41,77 @@
...
+                // TODO: Find alternative
+                //requestPermissions(new String[] {Manifest.permission.ACCESS_FINE_LOCATION}, PERMISSIONS_FINE_LOCATION);
+            }
+        }
+    }
 
+    private void getCoordinates() {
+        fusedLocationProviderClient = LocationServices.getFusedLocationProviderClient(this);
+
+        if (ActivityCompat.checkSelfPermission(this, Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED) {
```
Geno22, line 93: I find this if condition very long but this is just picking

app/src/main/java/com/github/houseorganizer/houseorganizer/house/HouseSelectionActivity.java
```diff
@@ -137,9 +202,21 @@ protected void onStop() {
         adapter.stopListening();
     }
 
+    @Override
+    public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions, @NonNull int[] grantResults) {
+        super.onRequestPermissionsResult(requestCode, permissions, grantResults);
```
Geno22, line 207: I am sorry but I don't understand what this function do ? Does it simply show if the permission is not granted ?

Geno22 (changes requested):
Thank you for the gps researches and advancements ! I made a few comments but I will approve quickly

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/HouseSelectionActivityTest.java
aindreias: That's a good question; the offline screen also uses some sort of permission

app/src/main/java/com/github/houseorganizer/houseorganizer/house/HouseSelectionActivity.java
sansive: Yes, it informs the user wether the requested permission has been given or not

aindreias (approved):
Nice work, looking forward to this feature. Not much to add besides what Florian already commented.

app/src/main/java/com/github/houseorganizer/houseorganizer/house/HouseSelectionActivity.java
sansive: True, I don't think it is necesary. I will remove it

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/HouseSelectionActivityTest.java
sansive: I think so, but I haven't still been able to prevent the pop up from appearing

Geno22 (approved):
Thanks for the changes !

PR #233, opened 2022-05-03
Feature/location

Geno22:
Added the house selection based on the current phone location.


app/src/androidTest/java/com/github/houseorganizer/houseorganizer/LocationHelperTests.java
```diff
@@ -0,0 +1,47 @@
...
+    public void calculateDistanceWorks(){
+        double lon1 = 10;
+        double lat1 = 5;
+        double lon2 = 20;
+        double lat2 = 9;
+        double result1 = 1189;
+        double result2 = 1190;
+        double out = LocationHelpers.calculateDistance(lon1, lat1,
+                lon2, lat2);
+        assertThat(out > result1 && out < result2, is(true));
```
Arkadyus, line 34: You can use assertEquals with an epsilon for double equality (https://stackoverflow.com/questions/35377956/assertequals-precision)

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/LocationHouseOnMainScreenTest.java
```diff
@@ -0,0 +1,83 @@
...
+        AtomicReference<FusedLocationProviderClient> locClient = new AtomicReference<>();
+        testRule.getScenario().onActivity(a -> locClient.set(a.fusedLocationClient));
+        Task<Void> t = locClient.get().setMockMode(true);
+        t.continueWithTask(r -> locClient.get().setMockLocation(location));
+        Tasks.await(t);
+        testRule.getScenario().recreate();*/
+        //testRule.getScenario().onActivity(a -> a.startActivity(intentFromLogin));
+        //startActivity(getApplicationContext(), intentFromLogin, null);
+        onView(withId(R.id.info_imageButton)).perform(click());
+        onView(withId(R.id.notesTextView)).check(matches(withText(FirebaseTestsHelper.TEST_HOUSEHOLD_DESC[0])));
```
Arkadyus, line 80: is this test still meaningful with all the commented out code ? Also another issue is that before we implemented locations it would already be house 0 by default, this test might pass even if the implementation doesnt actually work

app/src/main/java/com/github/houseorganizer/houseorganizer/house/HouseSelectionActivity.java
```diff
@@ -78,7 +77,7 @@ protected void onCreate(Bundle savedInstanceState) {
 
     private void requestPermission() {
         if (ActivityCompat.checkSelfPermission(this, Manifest.permission.ACCESS_FINE_LOCATION) != PackageManager.PERMISSION_GRANTED) {
-            // Comment next line to pass the tests, GrantPermissionRule doesn't prevent the pop up from appearomment next line to pass the tests, GrantPermissionRule doesn't prevent the pop up from appear
+            // Comment next line to pass the tests, GrantPermissionRule doesn't prevent the pop up from appearing
             // TODO: Find alternative
             //requestPermissions(new String[] {Manifest.permission.ACCESS_FINE_LOCATION}, PERMISSIONS_FINE_LOCATION);
         }
```
Arkadyus, line 83: is leaving this commented out fine ? does it just assume it has permission or how does it work since we dont request it

app/src/main/java/com/github/houseorganizer/houseorganizer/location/LocationHelpers.java
```diff
@@ -0,0 +1,98 @@
...
+                + Math.cos(lat1) * Math.cos(lat2)
+                * Math.pow(Math.sin(dlon / 2),2);
+
+        temp = 2 * Math.asin(Math.sqrt(temp));
+
+        // Radius of earth in kilometers
+        double r = 6371;
+
+        // calculate the result
+        return(temp * r);
```
Arkadyus, line 59: since we only care about relative distance, you might be able to remove the multiplication by r

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/MainScreenActivity.java
```diff
@@ -70,8 +85,16 @@ protected void onCreate(Bundle savedInstanceState) {
 
         mUser = FirebaseAuth.getInstance().getCurrentUser();
         db = FirebaseFirestore.getInstance();
+        fusedLocationClient = LocationServices.getFusedLocationProviderClient(this);
+
+        loadHouse = getIntent().hasExtra("LoadHouse");
+
+        if(!loadHouse) loadData();
+        if(loadHouse && LocationHelpers.checkLocationPermission(this, this)){ // TODO find a way for not having 2 "this"
```
Arkadyus, line 93: you might be able to call Activity.getContext or something similar, that way you only need one "this"

Arkadyus (changes requested):
Looks good ! Just a few questions about comments and code along with some tips

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/LocationHelperTests.java
Geno22: Great thanks !

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/LocationHouseOnMainScreenTest.java
Geno22: Yes I left it like that for another sprint because I couldn't figure out how to use mock locations. So it is not meaningful I can comment the whole class if you think it is better

app/src/main/java/com/github/houseorganizer/houseorganizer/house/HouseSelectionActivity.java
Geno22: I did not touch this part that was done by Sandra last sprint
I just changed the comment because it was repeated
This is used for adding the current house location when creating one but it is not functionnal yet

app/src/main/java/com/github/houseorganizer/houseorganizer/location/LocationHelpers.java
Geno22: It could be omitted but as the maths functions are approximate I am afraid the difference will be in the error margin without the mutliplication
What do you think ?

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/MainScreenActivity.java
Geno22: Yes done thanks !

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/LocationHouseOnMainScreenTest.java
Arkadyus: i see, thats fine by me like that then

app/src/main/java/com/github/houseorganizer/houseorganizer/house/HouseSelectionActivity.java
Arkadyus: oh okay no problem then

app/src/main/java/com/github/houseorganizer/houseorganizer/location/LocationHelpers.java
Arkadyus: thats also true, you can leave it like that since it was mostly me being nitpicky to start with

Arkadyus (approved):
This looks good to me, nice addition !

app/src/main/java/com/github/houseorganizer/houseorganizer/location/LocationHelpers.java
```diff
@@ -0,0 +1,98 @@
...
+        double dlon = lon2 - lon1;
+        double dlat = lat2 - lat1;
+        double temp = Math.pow(Math.sin(dlat / 2), 2)
+                + Math.cos(lat1) * Math.cos(lat2)
+                * Math.pow(Math.sin(dlon / 2),2);
+
+        temp = 2 * Math.asin(Math.sqrt(temp));
+
+        // Radius of earth in kilometers
+        double r = 6371;
```
aindreias, line 56: Perhaps this one could be saved as a constant?

aindreias (approved):
Nice work! I left a comment but it's 100% up to you if you want to introduce a constant or not for that case.

app/src/main/java/com/github/houseorganizer/houseorganizer/location/LocationHelpers.java
Geno22: Yes this woud be cleaner nice catch !

PR #234, opened 2022-05-04
Bug/task list

aindreias:
**Bug fixes | Updates**
- changing a task's title no longer crashes the app
- a task list is now attached to new households
- deleting a task removes the task pointer associated to it in the metadata file
- when the house is deleted the task list associated to it is deleted, as well

**Additional work**
- task assignees are now linked to firebase; you can assign members of the household to given tasks
- added missing tests for : adding & removing tasks, task assignees' database reflection
- `FirestoreTaskTest`: fix `metadataRef()` s.t. other test classes using it do not get a `NullPointerException`


Arkadyus:
Now that a user can remove themselves from a household, how would the assignees work along with that (say if someone who has a task assigned leaves the household where that task is), does it crash ? does the assignee stay/gets removed ?

aindreias:
> Now that a user can remove themselves from a household, how would the assignees work along with that (say if someone who has a task assigned leaves the household where that task is), does it crash ? does the assignee stay/gets removed ?

The user is not removed as an assignee, with that being said, the assignee adapter uses the freshest list of member emails from firebase. What this means is that the removed user's email will not be visible in-app, and remains as an "artifact" in the "assignees" array of the tasks they were assigned to - it will not cause any crashes in the app. I plan to remove this reference either after Luca's PR is merged, or on Friday afternoon (i.e. very early next sprint)

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/TaskListActivityTest.java
```diff
@@ -109,30 +114,47 @@ public void assigneePopUpIsDisplayed() {
         onView(withText(R.string.assignees_button)).perform(click());
 
         // This line will be changed to the user of the household
```
Geno22, line 116: You might be able to remove this comment now ^^

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/TaskListActivityTest.java
```diff
@@ -109,30 +114,47 @@ public void assigneePopUpIsDisplayed() {
...
+        changeTitleOrDesc(false);
+    }
+
+    @Test /* DB: unchanged */
+    public void changingTitleWorks() throws ExecutionException, InterruptedException {
+        changeTitleOrDesc(true);
+    }
+
+    private void changeTitleOrDesc(boolean changeTitle) throws ExecutionException, InterruptedException {
+        int inputId = changeTitle ? R.id.task_title_input : R.id.task_description_input;
```
Geno22, line 138: Clever way to combine tests well done !

app/src/main/java/com/github/houseorganizer/houseorganizer/house/EditHousehold.java
```diff
@@ -218,7 +220,27 @@ public void deleteGroceryList(View view) {
...
+
+                    DocumentSnapshot metadataSnap = docRefList.getDocuments().get(0);
+
+                    List<DocumentReference> taskPtrs = (ArrayList<DocumentReference>)
+                            metadataSnap.getData().getOrDefault("task-ptrs", new ArrayList<>());
+
+                    assert taskPtrs != null;
+                    taskPtrs.forEach(taskDocRef -> taskDocRef.delete().addOnFailureListener(tlDeletionFailed));
+
+                    metadataSnap.getReference().delete().addOnFailureListener(tlDeletionFailed);
```
Geno22, line 241: Is it a problem if the metadata is deleted before the taskPtr ? If not there is no problem but otherwise it might be better to delete it in the TaskPtr onFailureListener

Geno22 (changes requested):
Thanks a lot for those fixes ! I just added a comment about one thing

app/src/main/java/com/github/houseorganizer/houseorganizer/house/EditHousehold.java
aindreias: There is one case where we'd have a problem: if there's a taskPtr that can't be deleted, but we deleted the metadata, we have no idea which one it could have been. We have like 3 solutions:
-  schedule a periodical cleanup of non-linked tasks from task_dump
- log the taskPtrs that couldn't be deleted in a special file on firebase & have someone delete them manually
- delete the metadata if and only if all the taskPtrs were successfully deleted

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/TaskListActivityTest.java
aindreias: Thank you!

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/TaskListActivityTest.java
aindreias: Thanks for the catch! Will remove it soon 👍

app/src/main/java/com/github/houseorganizer/houseorganizer/house/EditHousehold.java
Geno22: Thanks for the reflexion and the fix !

Geno22 (approved):

Oxydeas (approved):
Nothing much to say there. Thanks for the bug fix.

PR #238, opened 2022-05-05
make the QR code button transparent

Arkadyus:
This makes the QR code button transparent as it used to have a white background instead (would show if the phone has dark mode by default)


aindreias (approved):
Perfect, thanks for thinking of dark mode users!

Geno22 (approved):
Good catch

PR #247, opened 2022-05-05
Bug/tests

sansive:
## Task
* Add use of IdlingResources to fix the problem related to firebase tasks that take too long
* Add before method on all test classes to dismiss alert dialog in the middle of tests

## To-Do
Remove thread.sleeps on the code for more reliability


aindreias (approved):
Nothing much to say for this one. Thanks a lot for your work! Looking forward to ultimately having sleep-free tests 👍

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/TaskListOnMainScreenTest.java
```diff
@@ -71,6 +75,9 @@ public static void signOut() {
     @Before
     public void forceTaskView() throws InterruptedException {
         Thread.sleep(2000); // no longer necessary to perform clicks
+
+        Context context = InstrumentationRegistry.getInstrumentation().getTargetContext();
+        context.sendBroadcast(new Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS));
```
Defibrilator, line 80: Could extract this to FirebaseTestHelpers

Defibrilator (changes requested):
Nice work thank you ! Just a little extraction and its good

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/TaskListOnMainScreenTest.java
Defibrilator: ah I see

Defibrilator (approved):

PR #255, opened 2022-05-06
Fixed location bug

Geno22:


aindreias (approved):
Thanks for the fix!

Defibrilator (approved):
Thanks for the fix !

PR #258, opened 2022-05-10
Update assignee arrays when a user is removed

aindreias:
What it says on the tin: this PR will keep our database clean by removing task assignments to recently deleted users of a household.


Arkadyus (approved):
Great, thanks for taking care of this !

PR #260, opened 2022-05-11
Feature/location

Geno22:
Added address translation so a user only has to enter its address instead of coordinates


app/src/main/java/com/github/houseorganizer/houseorganizer/house/CreateHouseholdActivity.java
```diff
@@ -36,23 +50,107 @@ protected void onCreate(Bundle savedInstanceState) {
...
+        }else{
+            try {
+                Location location = new Location("");
+
+                List<Address> geoResults = geocoder.getFromLocationName(address, 1);
+                int count = 0;
+                while (geoResults.size()==0 && count < 10) {
+                    geoResults = geocoder.getFromLocationName(address, 1);
+                    ++count;
+                }
```
Arkadyus, line 72: does getFromLocationName need to run multiple times to find a correct location sometimes ? how is that ? I can imagine it can find more than one location per address but wouldnt it return all of them at once ?
Also this one looks deprecated : https://developer.android.com/reference/android/location/Geocoder#getFromLocationName(java.lang.String,%20int,%20double,%20double,%20double,%20double)
using the one with the listener instead shouldnt be too hard

Arkadyus (changes requested):
Everything looks good to me, just one little question/change to make

app/src/main/java/com/github/houseorganizer/houseorganizer/house/CreateHouseholdActivity.java
Geno22: For the first part I found that sometimes for no reason it just returns null or an empty list, so this allows to ping it more than once to get an answer without having to push the button a second time.
For the second part it is deprecated for android 13 which is still in beta and the listener doesn't exist for the version we are using, even though I agree it would be way better

Arkadyus: Oh I see, that's my bad then, thanks for clarifying !

Arkadyus (approved):
Everything looks fine now, thanks !

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivityTest.java
```diff
@@ -69,19 +81,64 @@ public void setupHouseholds() throws ExecutionException, InterruptedException {
 
         Context context = InstrumentationRegistry.getInstrumentation().getTargetContext();
         context.sendBroadcast(new Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS));
+        createHouseholdRule.getScenario().onActivity(a -> decorView = a.getWindow().getDecorView());
```
Oxydeas, line 84: What is the purpose of this ?

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivityTest.java
```diff
@@ -69,19 +81,64 @@ public void setupHouseholds() throws ExecutionException, InterruptedException {
...
-                typeText("45"), closeSoftKeyboard());
-        onView(withId(R.id.editTextLongitude)).perform(click(),
-                typeText("45"), closeSoftKeyboard());
+                typeText(houseName), closeSoftKeyboard());
+        onView(withId(R.id.editTextAddress)).perform(
+                typeText("EPFL, Lausanne"), closeSoftKeyboard());
         onView(withId(R.id.submitHouseholdButton)).perform(click());
+        Thread.sleep(4000);
+        onView(withText(R.string.confirm)).perform(click());
+        Thread.sleep(1000);
```
Oxydeas, line 99: @sansive might want to get rid of this thread sleeps

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivityTest.java
```diff
@@ -69,19 +81,64 @@ public void setupHouseholds() throws ExecutionException, InterruptedException {
 
         Context context = InstrumentationRegistry.getInstrumentation().getTargetContext();
         context.sendBroadcast(new Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS));
+        createHouseholdRule.getScenario().onActivity(a -> decorView = a.getWindow().getDecorView());
     }
 
     @Test
     public void createHouseholdWorks() throws ExecutionException, InterruptedException {
+
+        final String houseName = "MyHouse";
```
Oxydeas, line 90: Create a variable at the start of the file such that you can reuse it

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivityTest.java
```diff
@@ -69,19 +81,64 @@ public void setupHouseholds() throws ExecutionException, InterruptedException {
...
+        assertEquals(house.size(), 1);
+    }
+
+    public void houseCreatedWithRightCoordinates() throws ExecutionException, InterruptedException {
+        onView(withId(R.id.editTextHouseholdName)).perform(click(),
+                typeText("MyHouse"), closeSoftKeyboard());
+        onView(withId(R.id.editTextAddress)).perform(
+                typeText("EPFL, Lausanne"), closeSoftKeyboard());
+        onView(withId(R.id.submitHouseholdButton)).perform(click());
+        Map<String, Object> house = FirebaseTestsHelper.fetchHouseholdData("MyHouse", db);
```
Oxydeas, line 114: Reuse variable above

app/src/main/java/com/github/houseorganizer/houseorganizer/house/CreateHouseholdActivity.java
```diff
@@ -36,23 +50,107 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
-    public void submitHouseholdToFirestore(View view){
+    public void createHouseholdButtonPressed(View view){
         TextView houseHoldNameView = findViewById(R.id.editTextHouseholdName);
-        TextView latitudeView = findViewById(R.id.editTextLatitude);
-        TextView longitudeView = findViewById(R.id.editTextLongitude);
+        TextView addressView = findViewById(R.id.editTextAddress);
+
+        if(houseHoldNameView.getText().toString().equals("") || addressView.getText().toString().equals("")){
+            Toast.makeText(this, R.string.address_fill_fields, Toast.LENGTH_LONG).show();
```
Oxydeas, line 137: Do you need long length for toast ?

app/src/main/java/com/github/houseorganizer/houseorganizer/house/CreateHouseholdActivity.java
```diff
@@ -36,23 +50,107 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            return;
+        }
+
+        String houseName = houseHoldNameView.getText().toString();
+        String address = addressView.getText().toString();
+
+        Location loc = getCoordinatesFromAddress(address);
+        if(loc == null){
+            Toast.makeText(this, R.string.address_error, Toast.LENGTH_LONG).show();
```
Oxydeas, line 146: Do you need long length for toast ?

app/src/main/java/com/github/houseorganizer/houseorganizer/house/CreateHouseholdActivity.java
```diff
@@ -79,7 +177,7 @@ private void signalFailure(Exception exception, View v) {
         startActivity(intent);
     }
 
-    private Map<String, Object> createHousehold(CharSequence houseHoldName, int lat, int lon) {
+    private Map<String, Object> createHousehold(CharSequence houseHoldName, double lat, double lon) {
```
Oxydeas, line 180: That's better than integer coordinates 😆

Oxydeas (changes requested):
Overall pretty good. Some questions and changes here and there but should take 10 minutes to fix. Great job.

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivityTest.java
Geno22: Oh yes don't need it anymore good catch

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivityTest.java
Geno22: the first one is special as we have to wait for the location to be retrieved but we can see if @sansive has a solution

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivityTest.java
Geno22: Yes good idea correcting it !

app/src/main/java/com/github/houseorganizer/houseorganizer/house/CreateHouseholdActivity.java
Geno22: I am not sure what it represents it is the time or the length ? in any case it displays well for now

Oxydeas: Ok then go with it 👍

app/src/main/java/com/github/houseorganizer/houseorganizer/house/CreateHouseholdActivity.java
Geno22: don't you like your position +- 100km ? xD

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivityTest.java
```diff
@@ -75,15 +85,57 @@ public void setupHouseholds() throws ExecutionException, InterruptedException {
...
+    public void emptyFieldsDoesNotCreateHouse() throws InterruptedException, ExecutionException {
+
+
+        Task<QuerySnapshot> t0 = db.collection("households").get();
+        onView(withId(R.id.submitHouseholdButton)).perform(click());
+
+        Task<QuerySnapshot> t1 = db.collection("households").get();
+
+        onView(withId(R.id.editTextAddress)).perform(typeText("address"), closeSoftKeyboard());
+        onView(withId(R.id.submitHouseholdButton)).perform(click());
```
Defibrilator, line 126: do you need line 122 ? since the second part of the if line 136 of CreateHouseholdActivity is not checked. Or you could also check when you input only the name and not the address.

Defibrilator (dismissed):
Good work ! two small comments.

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivityTest.java
sansive: Thanks for the mention, I'll look into it

Oxydeas (approved):
Thanks for the fixes

app/src/main/res/layout/activity_create_household.xml
```diff
@@ -58,7 +60,7 @@
         android:id="@+id/submitHouseholdButton"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
-        android:onClick="submitHouseholdToFirestore"
+        android:onClick="createHouseholdButtonPressed"
```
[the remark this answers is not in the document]
Geno22: Yes and with the right name I find it cleaner ^^

app/src/androidTest/java/com/github/houseorganizer/houseorganizer/CreateHouseholdActivityTest.java
Geno22: I do not really know how to remove text from the bar so I figured this tests enough do you think I should do this other case ?

PR #262, opened 2022-05-11
clean attachment name and delete attachment when deleting event

Arkadyus:
changed the name of the events attachments so they are simpler and make more sense
now deletes the linked attachment when deleting an event


Geno22 (approved):
Thanks !

aindreias (approved):
Well done, thanks for the changes!

PR #287, opened 2022-05-12
Feature/billsharer

Defibrilator:
Added the balance activity with its logic and the possibility to remove debts, fixed a couple of things such as initResidents and other linked bugs.

test coverage is going to drop again but my main and only task for next week is testing.


app/src/main/java/com/github/houseorganizer/houseorganizer/billsharer/Billsharer.java
```diff
@@ -153,6 +200,19 @@ public void removeExpense(Expense expense) {
         updateExpenses();
```
sansive, line 200: Shouldn't we refreshBalances after adding an expense too?

Defibrilator: No I do it when removing a debt but otherwise it will be refreshed when opening the balance tab

app/src/main/java/com/github/houseorganizer/houseorganizer/billsharer/DebtAdapter.java
```diff
@@ -0,0 +1,75 @@
...
+        return billsharer;
+    }
+
+    public void setBillsharer(Billsharer billsharer) {
+        this.billsharer = billsharer;
+    }
+
+    public void removeDebt(Context ctx, Debt debt, int pos) {
+        new AlertDialog.Builder(ctx)
+                .setTitle("Confirm payment of this debt ?")
```
aindreias, line 34: Can you extract this string?

aindreias (changes requested):
Nothing much to say, thanks for the feature! Just 1 string that you can extract & then I'm approving.

app/src/main/java/com/github/houseorganizer/houseorganizer/billsharer/Billsharer.java
sansive: Okey, just so I understand, why removing a debt yes and adding an expense no?

sansive (approved):
Good work!

aindreias (approved):

PR #292, opened 2022-05-13
fix expenses crashing because of nav_bar

Arkadyus:
This fixes the bug where the navigation bar would make the expenses screen crash upon trying to enter it
There is another known crash reason being trying to open the billsharer on a house that was created before the billsharer was implemented (as they dont have the firebase collections)


Oxydeas (approved):
Thanks for the fix

Geno22 (approved):
Thank for the fix !

PR #294, opened 2022-05-15
Exclude interfaces from code climate

sansive:
## Task
Exclude interfaces from code climate

### DIFF
--- a/app/src/main/java/com/github/houseorganizer/houseorganizer/panels/login/RegisterEmail.java
+++ b/app/src/main/java/com/github/houseorganizer/houseorganizer/panels/login/RegisterEmail.java
@@ -30,21 +30,27 @@ public class RegisterEmail extends AppCompatActivity {
 
     private FirebaseAuth mAuth;
     private boolean isEmailAlreadyUsed = false;
-    private String email;
-    private String password;
-    private String confPassword;
-    private TextView error_field;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
         setContentView(R.layout.activity_register_email);
 
-        initializeVariables();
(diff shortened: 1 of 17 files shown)


Oxydeas (approved):

Geno22 (approved):
Great let's hope it bumps up the coverage !

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/login/RegisterEmail.java
```diff
@@ -62,11 +62,18 @@ protected void onCreate(Bundle savedInstanceState) {
         );
     }
 
-
+    private void initializeVariables() {
+        mAuth = FirebaseAuth.getInstance();
+        isEmailAlreadyUsed = false;
+        email = ((EditText) findViewById(R.id.reg_enter_email)).getText().toString();
+        password = ((EditText) findViewById(R.id.reg_enter_password)).getText().toString();
+        confPassword = ((EditText) findViewById(R.id.reg_confirm_password)).getText().toString();
```
Geno22, line 70: I I think the elements that are in the fields are supposed to be retrieved only when clicking the button, here you take them at the beginning when they are empty I think

Geno22 (approved):
Thanks !

sansive:
@Oxydeas I pushed some new changes if you want to take a look at them before merging

Geno22 (commented):
According to codeclimate the total coverage is still 64%...
I don't get where it comes from

Oxydeas (approved):

sansive:
> According to codeclimate the total coverage is still 64%... I don't get where it comes from

Yes, that's true. I have posted a message on Github discussions to see if we can get any help

PR #296, opened 2022-05-17
fix billsharer crashing

Defibrilator:
Balance would crash because it did not recognize the nav bar id


Geno22:
It seems you just undid or did the same as @Arkadyus did in [this PR](https://github.com/House-Organizer/House-Organizer/pull/292).
You might wanna check with him

Defibrilator:
> It seems you just undid or did the same as @Arkadyus did in [this PR](https://github.com/House-Organizer/House-Organizer/pull/292). You might wanna check with him

No there were missing changes in his PR, the balance activity was still crashing

PR #301, opened 2022-05-18
Feature/facebook sign in

sansive:
## Task
* Implement Facebook log in

## Task
* Implement tests checking the Facebook Sign In button


sansive:
Skiped check from gitguardian as it doesn't matter to have those facebook keys exposed. The only key that shouldn't be shared is the "Secret app key" which haven't been shared.

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/login/LoginActivity.java
```diff
@@ -88,11 +108,70 @@ protected void onCreate(Bundle savedInstanceState) {
             }
     );
 
+    private void initializeFacebookLogIn() {
+        mCallbackManager = CallbackManager.Factory.create();
+        LoginManager.getInstance().logInWithReadPermissions(this, Arrays.asList("public_profile"));
+        LoginManager.getInstance().registerCallback(mCallbackManager, new FacebookCallback<LoginResult>() {
+            @Override
+            public void onSuccess(LoginResult loginResult) {
+                Log.d(getString(R.string.tag_login_activity), "facebook:onSuccess:" + loginResult);
```
Arkadyus, line 117: we should probably find a convention for when to use logs and when not to, i think until now we havent logged successful actions

Arkadyus (changes requested):
Everything looks good ! just a few questions about some small things

PR #302, opened 2022-05-19
Fix color themes' compatibility with dark mode

aindreias:
All 3 app themes are now compatible with dark mode. I also applied the change mentioned in our Thursday meeting (redirecting users to SettingsActivity instead of MainScreenActivity once the user changes either the app language or theme)


sansive (approved):
Everything looks good to me! Nice work!

Geno22 (approved):
Thanks for the fix !

PR #311, opened 2022-05-25
Anon users fix

aindreias:
**What's new**
- anon users are now assigned a fake email upon sign-in, so that they can use the app as any other email-registered user;
- when anon users sign out, they will be deleted from household member lists, as well as task assignee arrays

**Some notes for this / a future sprint**
- we could offer the same feature (data & account deletion) to any user who no longer wants to use HouseOrganizer
- Corner case: when a house whose only resident is deleted, the house itself is not deleted


Arkadyus (approved):
Everything looks good to me, nice to see this feature finally become something fully developed !

Oxydeas (approved):
I like your idea of offering the possibility to other users to remove their account. We could try it next week and if it leads to other issues we can just forget about it

PR #313, opened 2022-05-26
Feature/household image on info

Oxydeas:
This branch ads a button to display the image - if any - of the household


app/src/main/res/values-es-rES/strings.xml
```diff
@@ -133,6 +133,7 @@
     <string name="reminder">Recordatorio</string>
     <string name="reminder_message">No olvides tu tarea :</string>
     <string name="notify">Notificar</string>
+    <string name="show_image">TODO</string>
```
sansive, line 136: This would be "Mostrar imagen de la casa"

sansive (approved):
I tested it and it works perfectly! Great work! I added the spanish translation for the string g in case you want to remove the TODO now

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/info/InfoActivity.java
```diff
@@ -62,6 +65,23 @@ protected void onCreate(Bundle savedInstanceState) {
...
+            FirebaseStorage.getInstance()
+                    .getReference()
+                    .child("house_" + currentHouse.getId())
+                    .getDownloadUrl()
+                    .addOnCompleteListener(task -> {
+                if(task.isSuccessful()) {
+                    ImageHelper.showImagePopup(task.getResult(), view.getContext());
+                }
+                else {
+                    Toast.makeText(getApplicationContext(), "Could not find the attachment", Toast.LENGTH_SHORT).show();
```
Arkadyus, line 79: This might have to be put into a resource string so it can get translated, I just noticed its the same for attachments, we should look through our code to find these next week so its not that big a deal if its not done on this PR

Defibrilator (approved):
Works nicely, thank you for the addition !

Arkadyus (approved):
Great addition ! Just a comment but it can be handled next week

app/src/main/java/com/github/houseorganizer/houseorganizer/panels/info/InfoActivity.java
Oxydeas: I agree!

PR #316, opened 2022-05-26
Feature/edit shares

Defibrilator:
Added the possibility to specify shares when adding an expense.


app/src/main/java/com/github/houseorganizer/houseorganizer/billsharer/SharesAdapter.java
```diff
@@ -0,0 +1,83 @@
...
+        holder.amount.addTextChangedListener(new TextWatcher() {
+            @Override
+            public void beforeTextChanged(CharSequence s, int start, int count, int after) {}
+
+            @Override
+            public void onTextChanged(CharSequence s, int start, int before, int count) {}
+
+            @Override
+            public void afterTextChanged(Editable s) {
+                if (!holder.amount.getText().toString().equals("")) {
```
aindreias, line 60: Does this work correctly for shares >= 10? I'm asking because afterTextChanged is called after every typed character, if I'm not mistaken (found this while working on the TaskList adapters).

aindreias (changes requested):
Tested it out, everything looks great to me - just had 1 question, approving once it's clarified 👍

app/src/main/java/com/github/houseorganizer/houseorganizer/billsharer/SharesAdapter.java
aindreias: Wait, you're right, so in this case it's not an issue 👍 
On my side, I didn't change anything at the time, if you want to read the text only once, though, you could do so after the user confirms their input by pushing the positive alert dialog button.

aindreias (approved):
Nice work! And thanks for the explanation 👍

app/src/main/res/layout/share_editor.xml
```diff
@@ -0,0 +1,18 @@
...
+    android:orientation="vertical"
+    android:gravity="start">
+
+    <androidx.recyclerview.widget.RecyclerView
+        android:id="@+id/share_editor"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content" />
+
+    <EditText
+        android:layout_width="match_parent"
```
Geno22, line 15: There is a bunch of warnings here that are easily removable according to Cirrus. It might be nice to take it into account so we don't have to refactor it next week ^^.

Geno22 (approved):
Thanks for the changes !
However the coverage went down 2% 
Will you be able to bump it next week ?

PR #318, opened 2022-05-27
Fixed location denied crash

Geno22:


Arkadyus (approved):
what a simple fix wow

PR #319, opened 2022-05-30
document and clean up calendar classes along with a few other cleanups

Arkadyus:
This PR fully documents the public non override methods in the calendar classes
it also cleans up the public/private modifiers and makes them as restrictive as possible
NavBarHelpers was removed due to it being replaced by NavBarActivity
Removed a few unused methods too 
fixed a bug where the app crashed when returning from adding an image to an event/house without picking one
This PR should not change the behaviour of the app in any way, if it does please add a comment about it


Geno22 (approved):
Thanks for the documentation !

